PR #26, opened 2017-10-24
Add support for audio output streams!

mitchmindtree:
Allows for:

- Enumerating available devices
- Creating a new audio output stream with custom device, sample rate,
channels, sample format and buffer size.

The API works by creating a stream with some given state that is to live
on the audio thread along with a function that renders the state to an
audio buffer, returning the result of both. The user can also send
updates to the audio state from the GUI thread via the stream handle
using the `send` method.

A `simple_audio.rs` example has been added which renders a sine wave to
the default output stream. @freesig / @joshuabatty any chance either of
you chaps could test the example on macOS? It works on linux fine, just
(except `pause` and `unpause` haven't been implemented properly in
cpal), curious to hear if cpal is working sweet on macos atm.


mitchmindtree:
Known bugs:

- currently a `cpal::EventLoop` is spawned for each voice, whereas what should actually happen is that the `App` should run a single `cpal::EventLoop` once the first voice is spawned and then register each unique voice with that same event loop as they are spawned.

mitchmindtree:
Also, it would be cool to add some device event variants to the main event type which notified the user when an audio device is added or removed. I don't think there are any cross-platform APIs for this yet, but would possibly be something CPAL would be interested in adding support for.

mitchmindtree:
Going to merge as this works well enough for now and address the other issues mentioned above in follow up PRs :+1:

PR #35, opened 2017-10-29
 Add OSC support! Includes `Sender`, `Receiver` types and examples.

### LINKED ISSUE #34: OSC Support
[rosc](https://docs.rs/rosc) is a pure-rust encoding/decoding crate for OSC that should be perfect for this. That said, it provides a pretty low-level API in the sense that it requires users to manually setup UDP sockets and interacte with buffers of bytes (for encoding/decoding).

It would be nice if we could provide something that felt slightly higher level than this - something along the lines of `osc::Sender` and `osc::Receiver` types which abstract away some of the tedious byte handling.

mitchmindtree:
The aim with this implementation is to wrap the `std::net` and `rosc`
APIs in a way that allows for easily creating OSC channels in a simple
as possible, safe manner.

The Sender and Receiver types are the main two types and can be
constructed via the `osc::sender` and `osc::receiver` functions
respectively. Both types also have more flexible constructors which
allow for specifying alternative socket addresses and MTUs.

The APIs exposed by Sender and Receiver are similar to that of the
`std::sync::mpsc::{Sender, Receiver}` pair and should feel very familiar
to existing rust users.

The Receiver can receive in both blocking and non-blocking modes via
`recv` and `try_recv` respectively. It also provides `iter` and
`try_iter` methods for producing Iterators that yield received results.

The Sender can send any type that can be automatically converted into an
`osc::Packet` including `osc::Message`, `osc::Bundle` and
`(String, Vec<osc::Type>)` (shorthand for an `osc::Message`).

Examples
--------

The `osc_receiver.rs` and `osc_sender.rs` examples are designed to be
run as a pair. The `osc_sender.rs` example sends mouse events to the
`osc_receiver.rs` example which displays them in a big list in realtime.

TODO
----

- [ ] Look into enabling broadcast channels for Sender?
- [ ] Make examples more interesting once graphics support is added.
- [ ] Allow users to pass custom addresses to the examples via cmd line.

Closes #34.


JoshuaBatty:
OOOoohhh yes! nice work brother!

PR #59, opened 2018-02-17
Update for CPAL input stream support. Update conrod and glium. Add CHANGELOG.

mitchmindtree:
Would be nice to add an example demonstrating the new input stream suppot soon! Perhaps even a feedback example that passes audio directly from an input stream to an output stream.


mitchmindtree:
Published as **0.3.0**.

Issue #161, opened 2018-06-26
cargo run --example simple_window gives a window with background only


heyarne:
I just tried compiling the example to take a first look at the library. After installing rust via `brew` and running `cargo run --example simple_window` I received a window with an olive green background but no shapes on it. Mouse movement appeared to be logged as intended. I reran the command, this time including the `--release` flag and had the same result.

My system:
```
~/tmp/nannou master 
❯ cargo --version
cargo 1.26.0

~/tmp/nannou master 
❯ sw_vers
ProductName:	Mac OS X
ProductVersion:	10.13.5
BuildVersion:	17F77
```

This is on a MacBook Pro (13-inch, 2017, Two Thunderbolt 3 ports). If I can provide any additional information to debug this I'm happy to help.

mitchmindtree:
Hey thanks a lot for the issue!

This sounds like expected behaviour for the simple_window example - the aim for this one is to demonstrate the minimum code required for setting up a simple, empty window. Perhaps we could document this a bit better with a comment at the top! 

It sounds like maybe you are looking for the simple_draw example? It has some basic graphics that move around as you move the cursor. 

Hope this helps!

heyarne:
Oh, I see. I didn't even check the example source code yet, I was just following the README and the window that appeared didn't match the gif. It's indeed `simple_draw`. I'll open a pull request to fix this (this is also probably the more interesting example).

esimov:
First of all I think this is a nice project. But following the instructions from the readme file the gif is a bit misleading since the `simple_draw` example is totally different.

mitchmindtree:
Thanks for the feedback @esimov! @heyarne has already submitted a fix for this at #163 however travis seems to be failing to build docs for the serde crate on the beta and nightly channels. I'm just re-running travis to see if the issue has been fixed, but will probably end up merging this anyway as it's an important fix!

mitchmindtree:
Closed via #163.

PR #163, opened 2018-06-26
Use "simple_draw" in README

heyarne:
Makes sure the gif matches the example that is built. See #161


mitchmindtree:
Amazing, thanks so much for this! We're on a train without our lappies and wouldn't be able to fix it ourselves for a little while, this helps a lot :)

PR #183, opened 2018-07-23
Noc 6 01 seek

freesig:
- First example in NOC 06
- Tests for Vector2::angle()
- Example docs for Vector2::angle()

Not sure about my coding style but let me know if I should make any changes

### DIFF
--- a/examples/nature_of_code/chp_06_agents/6_01_seek.rs
+++ b/examples/nature_of_code/chp_06_agents/6_01_seek.rs
@@ -142,7 +142,7 @@ fn display(vehicle: &Vehicle, draw: &Draw) {
     } = vehicle;
     // Draw a triangle rotated in the direction of velocity
     // This calculation is wrong
-    let theta = velocity.angle() + 3.0 * (PI / 2.0);
+    let theta = (velocity.angle() + PI/2.0) * -1.0;
     let points = vec![
         pt3(0.0, -r * 2.0, 0.0),
         pt3(-r, r * 2.0, 0.0),
@@ -151,8 +151,9 @@ fn display(vehicle: &Vehicle, draw: &Draw) {
     draw.polygon()
         .points(points)
         .xy(*position)
-        .color(DARK_GREY)
-        //stroke(0);
-        //strokeWeight(1);
+        .color(Rgb::new(0.5, 0.5, 0.5))
+        //TODO add outline
+        //stroke(0),
+        //strokeWeight(1),
         .rotate(theta);
 }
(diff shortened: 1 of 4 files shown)


examples/nature_of_code/chp_06_agents/6_01_seek.rs
```diff
@@ -0,0 +1,159 @@
...
+fn main() {
+    nannou::app(model, event, view).run();
+}
+
+struct Model {
+    vehicle: Vehicle,
+}
+
+struct Vehicle {
+    position: Vector2,
```
mitchmindtree: Using `Point2` might be a little clearer as "point"s should generally refer to a specific location within space, while "vector"s describe some sort of translation or direction in space.

This used to matter a lot more when we used `cgmath` for the `Point2` and `Vector2` types whose behaviour were quite different (they had different trait implementations and different inherent methods), but since we switched to using our own it no longer really matters as `Point2` is simply a type alias for `Vector2` and they should both implement everything.

Also, I guess Processing uses PVector for everything, so maybe it's nicer to keep things simpler anyways - I'm not really sure!

Up to you whether or not you want to change this - just thought I'd mention it so you have the backstory n Point2 vs Vector2 :)

examples/nature_of_code/chp_06_agents/6_01_seek.rs
```diff
@@ -0,0 +1,159 @@
...
+    max_force: f32,
+    // Maximum speed
+    max_speed: f32,
+}
+
+impl Vehicle {
+    fn new(x: f32, y: f32) -> Self {
+        let position = Vector2::new(x, y);
+        let velocity = Vector2::new(0.0, -2.0);
+        let acceleration = Vector2::new(0.0, 0.0);
```
mitchmindtree: In the prelude, nannou provides a short-hand for constructing a Vector2/3/4:

```rust
let acceleration = Vector2::new(0.0, 0.0); 
```
is equivalent to
```rust
let acceleration = vec2(0.0, 0.0);
```

examples/nature_of_code/chp_06_agents/6_01_seek.rs
```diff
@@ -0,0 +1,159 @@
...
+    let draw = app.draw();
+    draw.background().color(WHITE);
+
+    let mouse = Vector2::new(app.mouse.x, app.mouse.y);
+
+    draw.ellipse()
+        // Missing Stroke
+        .x_y(mouse.x, mouse.y)
+        .radius(48.0)
+        .color(Rgb::new(0.78, 0.78, 0.78));
```
mitchmindtree: This builder line could also be written as `.rgb(0.78, 0.78, 0.78)`

examples/nature_of_code/chp_06_agents/6_01_seek.rs
freesig: No worries, Might just leave this one as it's consistent with the example

Issue #174, opened 2018-07-05
nannou-new new fail compiling without libcurl-devel


WnP:
build command for `curl-sys v0.4.5` required `libcurl-devel` install on the system otherwise it could ends up with some `openssl` version issue.

Logs:

```
error: failed to run custom build command for `curl-sys v0.4.5`
process didn't exit successfully: `/tmp/cargo-installsZaFk6/release/build/curl-sys-728ef4db7749783f/build-script-build` (exit code: 101)
--- stdout
Couldn't find libcurl from pkgconfig ("`\"pkg-config\" \"--libs\" \"--cflags\" \"libcurl\"` did not exit successfully: exit code: 1\n--- stderr\nPackage libcurl was not found in the pkg-config s
earch path.\nPerhaps you should add the directory containing `libcurl.pc\'\nto the PKG_CONFIG_PATH environment variable\nNo package \'libcurl\' found\n"), compiling it from source...

<…truncated output…>

--- stderr
fatal: not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
configure: WARNING: OpenSSL headers and library versions do not match.
/home/scl/.cargo/registry/src/github.com-1ecc6299db9ec823/curl-sys-0.4.5/curl/lib/vtls/openssl.c:124:22: error: static declaration of 'OpenSSL_version_num' follows non-static declaration
 static unsigned long OpenSSL_version_num(void)
                      ^~~~~~~~~~~~~~~~~~~
In file included from /usr/include/openssl/bio.h:69:0,
                 from /usr/include/openssl/evp.h:67,
                 from /usr/include/openssl/hmac.h:67,
                 from /usr/include/openssl/ssl.h:149,
                 from /home/scl/.cargo/registry/src/github.com-1ecc6299db9ec823/curl-sys-0.4.5/curl/lib/urldata.h:86,
                 from /home/scl/.cargo/registry/src/github.com-1ecc6299db9ec823/curl-sys-0.4.5/curl/lib/vtls/openssl.c:41:
/usr/include/openssl/crypto.h:340:15: note: previous declaration of 'OpenSSL_version_num' was here
 unsigned long OpenSSL_version_num(void);
               ^~~~~~~~~~~~~~~~~~~
make[2]: *** [Makefile:2023: vtls/libcurl_la-openssl.lo] Error 1
make[2]: *** Waiting for unfinished jobs....
[...]

PR #184, opened 2018-07-25
Update requirements in README

apiraino:
Hello,

I'm trying `nannou` for the first time (also new to Rust). I've bumped in a minor hiccup during the example compilation, I was missing a required library.

Also I've updated the links to the API documentation to current version (0.8.0)

Hope this helps

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ help out!
 - [**Goals**](#goals)
 - [**Why Rust?**](#why-rust)
 - [**Getting Started**](#getting-started)
-  - [**Install dependencies**](#install-deps)
+  - [**Platform-specific Setup**](#platform-specific-setup)
   - [**Install Rust**](#install-rust)
   - [**IDE Setup**](#ide-setup)
   - [**Nannou Examples**](#nannou-examples)
@@ -146,7 +146,7 @@ rustup component add rust-src rustfmt-preview rust-analysis
 Please see [this link](https://www.rust-lang.org/en-US/install.html) if you
 would like more information on the Rust installation process.
 
-### Platform-Specific Setup
+### Platform-specific Setup
 
 Depending on what OS you are running, you might require an extra step or two.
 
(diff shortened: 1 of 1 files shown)


apiraino:
perhaps we could include a mention to the requirements that @WnP mentioned in #174 ?

README.md
```diff
@@ -74,7 +75,7 @@ Nannou aims to provide easy, cross-platform access to the things that artists ne
     [listeners](https://doc.rust-lang.org/std/net/struct.TcpListener.html)
     via std.
 - **Device & I/O stream APIs**:
-  - [x] [Audio](https://docs.rs/nannou/0.5.2/nannou/app/struct.Audio.html).
+  - [x] [Audio](https://docs.rs/nannou/0.8.0/nannou/app/struct.Audio.html).
```
mitchmindtree: Thanks a lot for updating all these version updates!

I wonder if there's a way we can write these links so that we con't have to change these version numbers every time we update :thinking:

mitchmindtree:
**Re #174** - hmmm the strange thing is that I'm not sure all linux users are running into this. I don't remember having to make this step, but then again maybe I downloaded it via some other package using pacman! @apiraino did you have to take this step? If so perhaps you're right and we should add it.

README.md
apiraino: ok no problem. If you want to *always* pin the latest version of the documentation, we can simply pin that with `latest` in the path ;-)

apiraino:
re #174 - before updating the README, I'd like to hear from the reporter as I, too, can't reproduce that. I'd suggest to wait a bit and if we can amend this PR with additional information or can be merged as is.

thanks for looking at it!

apiraino:
@mitchmindtree here's an amended version of the README detailing the issue with curl lib dev package and LibresSSL that has been investigated in the other issue.

Safe for other fixes, I think we can merge this.

apiraino:
superseded by #188

PR #216, opened 2018-11-29
[WIP] Switch from OpenGL to Vulkan

### LINKED ISSUE #108: Investigate the possibility of switching to Vulkan
Requirements
------------------

- We must be able to support cross-platform compilation without requiring the user download any third-party libs, etc. This includes building and running on macOS and ultimately would include supporting ARM based processors like raspberry pi etc. It looks like some companies have been recently having some luck with apple-based platforms [here](https://www.khronos.org/news/press/vulkan-applications-enabled-on-apple-platforms).
- Must allow for freely creating commercial works without any platform-specific license caveats or 3rd-party non-free frameworks (don't know if this discludes Molten on macos anymore?).
- It must be easy enough to write the backend. If the process of integrating vulkano is at least almost as easy as it currently is to setup GL using glium then we should be OK. However, if we have to write our own entire graphics pipeline from scratch on top of very low-level vulkan support we probably would not have the time and would be better off sticking with OpenGL until something higher-level appears. Will have to investigate what kind of high-level APIs vulkan requires.
- Must work on stable rust.
- Must continue to work with a simple `cargo run`.
- Must actually simplify

Nice-to-haves
------------------

- We should be able to support existing GLSL code. It may not be worth switching from OpenGL if it means abandoning all the awesome shader code out there that already exists. However, if instead we could use some kind of rust-style shading language instead (like RLSL) to which glsl shaders could be ported then maybe it would be worth it.

After having a quick look at [vulkano](https://github.com/vulkano-rs/vulkano) it looks like it might already be close to ticking all these boxes today! We would want to test that the simple triangle example compiles and runs easily on each of the major platforms first (macos, windows, linux). If not, we will have to investigate how easy it is to automate the downloading/installation of any necessary 3rd party free tools etc via a `build.rs` script.

### LINKED ISSUE #208: Transition to Vulkan
This is a follow up to #108 with some actionable steps!

The general plan is to depend on [vulkano](https://github.com/vulkano-rs/vulkano) for its type-safe API and switch out every that we currently use glium for it.

This [Vulkan Tutorial](https://github.com/bwasty/vulkan-tutorial-rs) should be a useful reference.

- [x] Run some example locally to verify that a few essentials currently work on stable:
    - Basic triangle.
    - GLSL to SPIR-V.
    - Textures (images).
- [x] Switch from using the glutin window and glcontext to vulkano-win.
- [x] Expose `vulkano::swapchain::Surface` methods from `Window`.
- [x] Add method for enumerating vulkan physical devices.
- [x] Add window builder that allows for specifying the physical device to use for the window's swapchain.
- [x] [Construct swapchain](https://vulkano.rs/guide/swapchain-creation) for window. Investigate whether this should be shared or unique between multiple windows.
- [x] Add `LoopMode` that only `update`s when the next image is ready for a specific window. ~Should consider making this a builder method on a window so that a unique update may be used for each if desired?~ Can't do this as i would require adding type params to the `App`.
- [x] Recreate `Frame` type using [`swapchain::acquire_next_image`](https://vulkano.rs/guide/acquire-present) future.
- [x] Get `LoopMode::RefreshSync` run loop working with a cleared image.
- [x] Add in swapchain recreation on resize.
- [x] Multi-window support.
- [x] Update other `LoopMode`s.
- [x] Recreate swapchains on `LoopMode` switch.
- [x] A window builder method that allows the user to construct the swapchain themselves.
- [x] Create vulkan based `draw` backend to replace the glium one. Refer to [triangle example](https://github.com/vulkano-rs/vulkano-examples/blob/master/src/bin/triangle.rs) for handling buffers.
- [x] Switch from using the glium backend of conrod to a vulkano one.

mitchmindtree:
This switches the windowed-graphics API from OpenGL (via glium) to
Vulkan (via vulkano). You can read more about the motivations, thoughts
and discussion behind this switch at #208 and #108.

There are still a few items left to complete before this PR is ready:

- [x] Fix bug where view seems to be vertically flipped (probably in
  vertex mapping process).
- [x] Fix bug where alpha channels always seem opaque.
- [x] Add vulkan debug support. See the [debug.rs](https://github.com/vulkano-rs/vulkano-examples/blob/master/src/bin/debug.rs) vulkano example.
- [x] Add depth attachment to draw renderpass.
- [x] Update old loop modes for changes in the application loop.
- [x] Recreate swapchain on loop mode switch.
- [x] ~Fix bug where clearing an image that is then multisampled does not
  work.~ Turns out this is a bug in mesa. See vulkano-rs/vulkano#1123
- [x] ~Fix bug where resizing vk_image.rs example panics on macos.~ Old moltenvk bug.
- [x] ~Fix bug where simple_window.rs panics on close on macos.~ Old moltenvk bug.
- [x] Merge conrod vulkan backend and switch to it. PistonDevelopers/conrod#1204
    - [x] Switch use of `CpuAccessibleBuffer` to `ImmutableBuffer` for vertex buffer and move creation out of command loop.
    - [x] Add proper error handling to vulkano backend.
    - [x] Return immutable buffer creation futures from `draw`?
- [x] Change `draw` to only create one of the `depth`, `multisampled_color` and `multisampled_depth` images and share them between each framebuffer, rather than one per framebuffer as the same graphics pipeline is used to render to each framebuffer anyway.
- [x] Create different render pass for `draw` renderer for when there is no background.
- [x] Update README for switch.
- [x] Update CHANGELOG.
- [ ] Merge and publish vulkano-rs/vulkano#1117 or related fix so we can
  switch to crates.io dep.
- [ ] Automate build process for macos and windows where possible.

Closes #208.
Closes #108.

Future related work:

- Create dynamic `RenderPassBuilder` and `RenderPassDescription` types in `gpu` module.
- Expose subpass rendering option for `draw`module allowing it to be rendered via different render passes/framebuffers.


src/app.rs
```diff
@@ -43,16 +103,14 @@ use window::{self, Window};
 /// - **All OpenGL windows** for graphics and user input. Windows can be referenced via their IDs.
```
JoshuaBatty: Line 103, Should be Vulkan instead of OpenGL

mitchmindtree:
OK, going to merge this into branch `v0.9`!

Once we can confirm builds are working on each platform and any necessary upstream patches have been merged and published we'll do a PR for merging `v0.9` into `master` and publish it :+1: In the meantime I'm going to keep doing new dev on the `v0.9` branch.

PR #221, opened 2018-12-13
Add window-specific view and event user function support. Add app update user function.

mitchmindtree:
An `all_functions.rs` example has been added that shows all functions
that can be registered with the app and model.

All other examples have been updated to take advantage of the new event
functions.

Functions that update the model, including the existing App `event`
function, now provide the model via `&mut` rather than by value to be
returned.

@JoshuaBatty might wanna check out all the updated NoC examples, so much nicer!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,10 @@
 - The `view` function is now called separately for each frame for each window,
   rather than a single frame for all windows at once. The window a frame is
   associated with can be determined via `Frame::window_id`.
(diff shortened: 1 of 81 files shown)


JoshuaBatty:
This is amazing! Nannou just got a whole lot easier and approachable 👍 
About to pour over this epic diff for the next hour or so :)

JoshuaBatty:
Just finished looking over all this, these examples are looking SO much neater now!

PR #260, opened 2019-02-12
Prefer sRGB color formats when building swapchain

mitchmindtree:
See the vulkano `ColorSpace` documentation
[here](https://docs.rs/vulkano/0.11.1/vulkano/swapchain/enum.ColorSpace.html)
for the motivation behind this change.


mitchmindtree:
@freesig I've tested this on the desktop that was having colour issues and it seems to have fixed them! Could you test this locally and let me know if it fixes your colour issues? If so I'll go ahead and merge it into v0.9

freesig:
Should be fine to merge this 👍

PR #265, opened 2019-02-13
Fallback to `Fifo` present mode if `Mailbox` is unavailable

### LINKED ISSUE #262: Mac UnsupportedPresentMode
This loop mode in the `simple_ui.rs` is not supported by MoltenVK
`app.set_loop_mode(LoopMode::wait(3));`
If I try and run it I get 
```
thread 'main' panicked at 'failed to recreate swapchain for new `LoopMode`: UnsupportedPresentMode'
```

Is there an alternative low energy mode?

mitchmindtree:
Nannou will now fall back to the `Fifo` present mode if `Mailbox` is
not supported by the swapchain when switching the `Wait` or `Rate`
`LoopMode`s.

This should fix the issue on macos where certain examples that use the
`Wait` or `Rate` loop modes would `panic!` with an
`UnsupportedPresentMode` error.

Closes #262.

@freesig could you confirm that this fixes the `simple_ui.rs` example
for you? It's probably also worth checking that the `loop_mode.rs` works
now too as that example switches between all of the different loop modes
one at a time on keypress.

Closes #262

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -573,11 +573,11 @@ where
         if loop_mode != LoopMode::default() {
             let mut windows = app.windows.borrow_mut();
             for window in windows.values_mut() {
-                let min_image_count = window
+                let capabilities = window
                     .surface
                     .capabilities(window.swapchain.device().physical_device())
-                    .expect("failed to get surface capabilities")
-                    .min_image_count;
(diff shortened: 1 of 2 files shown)


freesig:
Yep this works

PR #272, opened 2019-02-16
Add `FramebufferObject` type to simplify working with framebuffers.

### LINKED ISSUE #214: FBO API
Wanted to jot down some ideas on the frame buffer API and some use cases in Nannou. 

A frame buffer is a pretty simple concept and used all the time in modern graphics programming, so it would be great to create a really easy to instantiate and use API around it. 

Below outlines the functionality that I have used framebuffers in the past for. 

**1. Allocate** 
This step should create a new FBO in memory and should at a minimum define the width and height of the object. Also, there should be optional builder methods for defining:
- ColorFormat -> RGB, RGBA, RGBAF32, ([Here](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glTexImage2D.xhtml) are a bunch of GL colour formats)
- Number of Samples (used for MSAA)
- Number of Render Layers (the number of render targets associated with the FBO)
- Depth Image (should the FBO also contain a depth buffer)
- Stencil Image (should the FBO contain a stencil buffer)  

**2. Draw Something Into the FBO**
I'm imaging this will need some sort of `begin()` and `end()` calls. For example call fbo.begin() and then draw the contents of a fragment shader onto the fbo, then call end() after the frag shader is finished. 

During this stage we also need a way of clearing the currently stored memory in the FBO to an RGBA colour. This is especially needed for doing alpha trail like effects, and also when you are doing ping pong buffering. For example sometimes a shader will work on the Green channel but not the Blue and Red channels so it's important for clear the FBO to Green before each update pass for the effect to work. 

**3. Draw the FBO** 
Pretty straight forward just drawing the FBO onto a window. I'm imagining this will inherit all of the geometry draw builder methods, like x_y(). w_h(), rotation() etc.... 
If there is only a single render layer then calling fbo.draw() should present that, otherwise there should be a way of specify which render layer you want to draw. 

Similarly there should also be methods for draw_stencil() and draw_depth().  

**4. Get a reference to the FBO's texture**
This is used a lot when passing a texture into a shader. You want to just pass a reference to the memory location of the FBO on the GPU for the shader to use. Just like above, should have optional way of retrieving which render layers texture you want, or if you want the stencil of depth textures. 

**5. Loading Data directly into the FBO**
This is useful when using the texture data from a FBO in a shader to drive variables. For example, lets say you want to define the starting XY positions for a particle system shader. You should be able to store the X position in the Red channel and Y position in the Green channel. Each pixel in the FBO would be associated with a different particle. You can then store the velocity in the Blue channel. This is just a single example use case obviously. Below is a screen shot of an FBO using this type of system with particle system shader.
screen shot 2018-11-22 at 11 32 29

That's all I can think of for now. @mitchmindtree @freesig ...... Am I missing anything here?

mitchmindtree:
Also includes an `Fbo` alias for the new `FramebufferObject` type.

The `ViewFramebuffer` type has been replaced with a
`ViewFramebufferObject` type along with a `ViewFbo` type alias.
This new type uses the `Fbo` internally, greatly simlifying the
implementation.

Closes #214.

cc @JoshuaBatty this should simplify all the post-processing stuff heaps!

### DIFF
--- a/examples/vulkan/vk_image.rs
+++ b/examples/vulkan/vk_image.rs
@@ -24,7 +24,7 @@ struct Model {
     render_pass: Arc<RenderPassAbstract + Send + Sync>,
     pipeline: Arc<GraphicsPipelineAbstract + Send + Sync>,
     vertex_buffer: Arc<CpuAccessibleBuffer<[Vertex]>>,
-    framebuffer: RefCell<ViewFramebuffer>,
+    view_fbo: RefCell<ViewFbo>,
     desciptor_set: Arc<DescriptorSet + Send + Sync>,
 }
 
@@ -141,13 +141,13 @@ fn model(app: &App) -> Model {
             .unwrap(),
     );
 
-    let framebuffer = RefCell::new(ViewFramebuffer::default());
+    let view_fbo = RefCell::new(ViewFbo::default());
 
     Model {
         render_pass,
         pipeline,
         vertex_buffer,
-        framebuffer,
+        view_fbo,
         desciptor_set,
     }
 }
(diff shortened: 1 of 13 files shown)


JoshuaBatty:
shits getting serious! :)

PR #274, opened 2019-02-17
Embrace the `vk` prefix. Re-export vulkano via a new `vk` module.

mitchmindtree:
Also removes the `gpu` module in favour of the new `vk` module.

Many vulkano items have been re-exported into the `vk` module as well to
avoid name repetition and ease access to common types. Items that would
have an ambiguous meaning if they were to be re-exported have been left
out of the re-export list.

Methods and fields on the `App` and `Window` types starting with `vulkan_`
have been renamed to start with `vk_`.

See the examples for how to update and rid of your boilerplate today!

### DIFF
--- a/examples/vulkan/vk_compute_shader.rs
+++ b/examples/vulkan/vk_compute_shader.rs
@@ -1,29 +1,18 @@
 extern crate nannou;
 
 use nannou::prelude::*;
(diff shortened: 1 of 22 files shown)


JoshuaBatty:
Hot damn!

PR #275, opened 2019-02-17
Add a `ViewportBuilder` type.

mitchmindtree:
This simplifies the viewport creation process by removing the need to
specify the origin and depth_range, which can use reasonable defaults in
the vast majority of cases.

The defaults can be retrieved via `ViewportBuilder::DEFAULT_ORIGIN` and
`ViewportBuilder::DEFAULT_DEPTH_RANGE`.


mitchmindtree:
Tested successfully locally, merging.

PR #286, opened 2019-03-30
Fix panic in nature of code CA example 7_04

mitchmindtree:
Currently writing a test that runs all examples one at a time locally
and noticed that this example was `panic!`ing with a subtraction
underflow error. I've changed it to use signed integers instead to avoid
`panic!`.

However, it still seems to behave a bit strangely. I can't see any of the
1-dimensional CA patterns until the rows reach the top of the window and
after that they start disappearing.

@JoshuaBatty I'll probs merge this PR as is, but after could you maybe
try running this to see if it's behaving as intended sometime? I wonder
if the odd behaviour is a side-effect of the change where the background
no longer automatically clears? I remember you mentioning you were doing
something to work around that a while back, maybe whatever that is isn't
needed anymore?


mitchmindtree:
Travis is failing due to a bug in rustc on the latest nightly. Going to merge and re-run travis on master later.

Issue #271, opened 2019-02-14
Issues with Vulkan image layout transitions


mitchmindtree:
One of the final blockers before landing v0.9 is the state of automatic layout transitions in vulkano. Before explaining the issue, it's important to understand 1. image layouts, 2. image memory barriers and 3. image layout transitions.

### Image Layouts

The [`ImageLayout`](https://docs.rs/vulkano/0.11.1/vulkano/image/enum.ImageLayout.html) type describes the layout of an image (e.g. `AttachmentImage`, `StorageImage` or `SwapchainImage` - anything that implements `ImageAccess`) in GPU memory at each stage within a vulkan program. The thing that determines what `ImageLayout` an image should have at each stage is the way it is used. *I believe this is specified by [`ImageUsage`](https://docs.rs/vulkano/0.11.1/vulkano/image/struct.ImageUsage.html), but there could be other factors involved in determining required layout - I'm unclear on this*. This is highlighted nicely by a passage from the Vulkan cookbook:

> For example, sampling data from images inside shaders may need them to be cached in such a
> way that neighbor texels are also neighbors in memory. But, writing data to images may be
> performed faster when a memory is laid out linearly. That's why image layouts were
> introduced in Vulkan. Each image usage has its own, designated layout.

You can find a list of all possible image layouts supported by the spec along with descriptions [here](https://www.khronos.org/registry/vulkan/specs/1.1-extensions/man/html/VkImageLayout.html). *Side note: Vulkano currently seems to enumerate only a subset of these in its [`ImageLayout`](https://docs.rs/vulkano/0.11.1/vulkano/image/enum.ImageLayout.html) type.*

### Image Memory Barriers and Layout Transitions

Let's say we have a `SwapchainImage` and we want to clear it with a single colour and then present it via the swapchain.

To clear the image, we can use the [`vkCmdClearColorImage`](https://www.khronos.org/registry/vulkan/specs/1.1-extensions/man/html/vkCmdClearColorImage.html) command (via the `clear_color_image` commandbuffer builder method in vulkano). The `vkCmdClearColorImage` operation requires that the image is in the `VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL` layout (`ImageLayout::TransferDstOptimal` in vulkano).

[...]

PR #287, opened 2019-03-30
Add test that runs every nannou example listed in the Cargo.toml

mitchmindtree:
This adds an example that runs each nannou example listed in the
Cargo.toml for a few seconds each, one at a time.

This is useful for finding `panic!`s that only occur at runtime, as
while `cargo test` does compile the examples, it does not try and
execute them.

The test has the `#[ignore]` attribute so that `cargo test` does not run
the example by default. This is because 1. the test takes a looong time
to run and 2. travis is likely incapable of running a lot of the tests
as it doesn't have a windowing system etc. As a result, to run the test
locally use `cargo test -- --ignored`. We should try to ensure we run
this at least once before publishing each major release.

Specifically, this test was created to help with testing for vulkano
image layout bugs (#271). cc @freesig just pinging you as you might
find this useful for this reason too!

### DIFF
--- a/tests/run_all_examples.rs
+++ b/tests/run_all_examples.rs
@@ -1,0 +1,38 @@
+// A simple little script that attempts to run every example listed within the `Cargo.toml`.
+//
+// This will fail if any of the examples `panic!`.
+#[test]
+#[ignore]
+fn test_run_all_examples() {
+    // Read the nannou cargo manifest to a `toml::Value`.
+    let manifest_dir = env!("CARGO_MANIFEST_DIR");
+    let manifest_path = std::path::Path::new(manifest_dir).join("Cargo").with_extension("toml");
+    let bytes = std::fs::read(&manifest_path).unwrap();
+    let toml: toml::Value = toml::from_slice(&bytes).unwrap();
+
+    // Find the `examples` table within the `toml::Value` to find all example names.
+    let examples = toml["example"].as_array().expect("failed to retrieve example array");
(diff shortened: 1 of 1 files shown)


mitchmindtree:
Travis seems to be failing due to a bug in the latest nightly (rustc overflowing its stack while building docs):

```
thread 'rustc' has overflowed its stack
fatal runtime error: stack overflow
error: Could not document `num-traits`.
Caused by:
  process didn't exit successfully: `rustdoc --crate-name num_traits /home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/num-traits-0.2.6/src/lib.rs --cap-lints allow --color always -o /home/travis/build/nannou-org/nannou/target/doc --cfg 'feature="default"' --cfg 'feature="std"' -L dependency=/home/travis/build/nannou-org/nannou/target/debug/deps --cfg has_i128` (signal: 6, SIGABRT: process abort signal)
warning: build failed, waiting for other jobs to finish...
error: build failed
The command "cargo doc -v" exited with 101.
```

mitchmindtree:
Here is the relevant issue rust-lang/rust#59502 

Going to merge this for now as stable and beta seem to be building fine.

mitchmindtree:
Just a heads up that using `cargo test -- --ignored` will actually also try to compile and run examples that have been ignored throughout the docs. The correct way to run the new example would be to use:
```
cargo test -- --ignored run_all_examples
```

PR #292, opened 2019-04-15
Add an `examples/laser` directory demonstrating laser streams from the new `lasy` crate.

mitchmindtree:
Recently I started a new crate for discovering and streaming to laser
DACs called [**lasy**][1]. This commit adds two examples demonstrating
how to work with both the RawStream and FrameStream APIs from the new
crate.

The **laser_raw_stream.rs** example demonstrates how to feed points to a
raw stream by drawing a single point at a position controlled by the
position of the mouse within the window.

The **laser_frame_stream.rs** example demonstrates how to feed points
to a frame stream by switching between different test patterns with the 1, 2,
3 and 4 keys.

Both examples could be improved in the future by targeting all lasers
detected (rather than just the first) and using the
`ether-dream-dac-emulator` crate to assist with creating a visualisation
of what the DAC is outputting. Currently there's a demo of how to
visualise an emulated DAC in the ether-dream repo, but it might be best
to move that example into its own configurable visualiser type in a separate
crate so that it is super easy to drop into laser examples or other
downstream laser code. 

Originally I started integrating `lasy` directly into nannou itself, but
was weary of adding the extra compile-time work when the vast majority
of users probably don't have access to lasers. We can reconsider if we
can find a good motivation to do so, but otherwise I think the new
examples already demonstrate a decent level of simplicity!

[1]: https://github.com/nannou-org/lasy

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,7 +38,7 @@ winit = "0.19"
 moltenvk_deps = "0.1"
 
 [dev-dependencies]
-lasy = { git = "https://github.com/nannou-org/lasy", branch = "master" }
+lasy = "0.2"
 
 # --------------- Nannou Examples
 [[example]]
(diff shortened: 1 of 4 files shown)


mitchmindtree:
Just added a **laser_frame_stream_gui.rs** example that demonstrates how to tweak the laser DAC stream settings along with some other useful params in real-time.

The examples now depend on the latest crates.io version of lasy rather than the git repo.

PR #294, opened 2019-04-24
Hotloading shader example

freesig:
This is a working example now. I think it's time to get some other eyes on it.
The shaders are in `examples/vulkan/shaders/hotload_vert.glsl` and `examples/vulkan/shaders/hotload_frag.glsl`.

Still todo:
- [ ] Get descriptor set info from spirv-reflect
- [ ] Run tests with different / changing inputs outputs and descriptor sets
- [x] General tidy up


mitchmindtree:
Nice work!!!

I'll do a proper review of this soon but probably won't get a chance til this weekend. Looks good at a quick glance though!

freesig:
I think I might actually make this into a crate so that it is easier to use in other places

mitchmindtree:
@freesig I just gave the `vk_hotload.rs` example a test on my linux lappy - the first file change worked perfectly but then any changes after that didn't seem to be registered (the `shader_msg` was not receiving any more messages). Are you seeing this on macos? If not, I wonder if its maybe something to do with a difference in behaviour from notify between macos and linux...

freesig:
I am not seeing this behavior on macos and you're probably correct. I'd say it's a different [event](https://docs.rs/notify/4.0.10/notify/enum.DebouncedEvent.html) on linux. I'll see if I can figure out what it might be but if you have a chance it could be good to clone [shade runner](https://github.com/freesig/shade_runner/).
Chane [this line](https://github.com/freesig/shade_runner/blob/6eca607ca3a167ad2c5aefdb50fcf611834d527d/src/watch.rs#L103) to:
```rust
let msg = notify_rx.recv_timeout(Duration::from_secs(1));
dbg!(msg);
if let Ok(notify::DebouncedEvent::Create(_)) = msg
```
and add this

```
[patch.crates-io]
shade_runner = { path = '../shade_runner' }
```
Then you should be able to see exactly what events are coming out.
Alternatively I could just make it any event that happens to that file.

freesig:
If the above does work it might be worth chucking a dbg on shader_msg and making sure your aren't getting double events. I'm not on macos but there's a chance `NotifyWrite` and `Write` both emit on linux I think.

mitchmindtree:
I just tried `cargo update` and `git pull freesig hot` but I'm getting the exact same behaviour (only updates on the very first edit). I'll check the notify examples first to make sure my vim edits/writes are actually generating events and then have a look at `shade_runner` in a little bit :+1:

freesig:
dam. Yeh have a go with the notify examples and let me know what events it gets then I'll update shade runner.

freesig:
also this is the setup I'm using fyi
```rust
let mut watcher: RecommendedWatcher =
Watcher::new(notify_tx, Duration::from_millis(50)).expect("failed to create watcher");
```

mitchmindtree:
OK I just ran `cargo update` and tried `cargo run --example vk_hotload` but am getting this `panic!` now instead:

```
thread 'main' panicked at 'failed to read src: Os { code: 21, kind: Other, message: "Is a directory" }', src/libcore/result.rs:997:5
stack backtrace:
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
             at src/libstd/sys/unix/backtrace/tracing/gcc_s.rs:39
   1: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:71
   2: std::panicking::default_hook::{{closure}}
             at src/libstd/sys_common/backtrace.rs:59
             at src/libstd/panicking.rs:197
   3: std::panicking::default_hook
             at src/libstd/panicking.rs:211
   4: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:474
   5: std::panicking::continue_panic_fmt
             at src/libstd/panicking.rs:381
   6: rust_begin_unwind
             at src/libstd/panicking.rs:308
   7: core::panicking::panic_fmt
             at src/libcore/panicking.rs:85
   8: core::result::unwrap_failed
             at /rustc/3de0106789468b211bcc3a25c09c0cf07119186d/src/libcore/macros.rs:18
   9: core::result::Result<T,E>::expect
             at /rustc/3de0106789468b211bcc3a25c09c0cf07119186d/src/libcore/result.rs:825
  10: shade_runner::compiler::compile
             at /home/mindtree/.cargo/git/checkouts/shade_runner-2c0d6a7afd3f6fe6/5840e83/src/compiler.rs:14
  11: shade_runner::load
             at /home/mindtree/.cargo/git/checkouts/shade_runner-2c0d6a7afd3f6fe6/5840e83/src/lib.rs:28
  12: shade_runner::watch::Loader::reload
             at /home/mindtree/.cargo/git/checkouts/shade_runner-2c0d6a7afd3f6fe6/5840e83/src/watch.rs:56
  13: shade_runner::watch::Loader::create
             at /home/mindtree/.cargo/git/checkouts/shade_runner-2c0d6a7afd3f6fe6/5840e83/src/watch.rs:51
  14: shade_runner::watch::create_watch
             at /home/mindtree/.cargo/git/checkouts/shade_runner-2c0d6a7afd3f6fe6/5840e83/src/watch.rs:99
  15: shade_runner::watch::Watch::new
             at /home/mindtree/.cargo/git/checkouts/shade_runner-2c0d6a7afd3f6fe6/5840e83/src/watch.rs:32
  16: vk_hotload::model
             at examples/vulkan/vk_hotload.rs:55
  17: nannou::app::Builder<M,E>::run
             at ./src/app.rs:600
  18: vk_hotload::main
             at examples/vulkan/vk_hotload.rs:30
  19: std::rt::lang_start::{{closure}}
             at /rustc/3de0106789468b211bcc3a25c09c0cf07119186d/src/libstd/rt.rs:64
  20: std::panicking::try::do_call
             at src/libstd/rt.rs:49
             at src/libstd/panicking.rs:293
  21: __rust_maybe_catch_panic
             at src/libpanic_unwind/lib.rs:87
  22: std::rt::lang_start_internal
             at src/libstd/panicking.rs:272
             at src/libstd/panic.rs:388
             at src/libstd/rt.rs:48
  23: std::rt::lang_start
             at /rustc/3de0106789468b211bcc3a25c09c0cf07119186d/src/libstd/rt.rs:64
  24: main
  25: __libc_start_main
  26: _start
```

mitchmindtree:
Also side note: I think once we get this working nicely it might be nice to do a follow-up PR where we:

1. Use two triangles to create a rect that fills the screen.
2. Add some uniforms or push constants for feeding in time, aspect ratio, mouse position, etc.

This way the example might be closer to a native, vulkan based version of shadertoy and make it a little more obvious why hot-loading is such a cool feature. Thoughts?

freesig:
Yeh that sounds like a good idea. 
I'm currently removing all the `unimplemented!` and `expects` in favor of `Errors`. I'll fix that panic aswell in the next commit.

freesig:
Ok shad_runner shoudn't panic for any reason now.
Also the above error should be fixed.
You need to pull latest and cargo update

freesig:
https://github.com/freesig/shade_runner/issues/13 for shader toy style example

examples/vulkan/vk_hotload.rs
```diff
@@ -0,0 +1,246 @@
...
+    let device = app.main_window().swapchain().device().clone();
+
+    // We now create a buffer that will store the shape of our triangle.
+    let vertex_buffer = {
+        let positions = [[-0.5, -0.25], [0.0, 0.5], [0.25, -0.1]];
+        let vertices = positions.iter().map(|&position| Vertex { position });
+        vk::CpuAccessibleBuffer::from_iter(device.clone(), vk::BufferUsage::all(), vertices)
+            .unwrap()
+    };
+    let nannou_root = std::env::current_dir().expect("failed to get root directory");
```
mitchmindtree: A safer way of getting this is to use the `CARGO_MANIFEST_DIR` env var, like this:
```rust
concat!(env!("CARGO_MANIFEST_DIR"), "examples/vulkan/shaders/hotload_vert.glsl");
```

I think this should give you a `&'static str` containing the full absolute path. This is a little more robust in the case that the user doesn't happen to be `cd`d into the nannou directory when they're running the example.

examples/vulkan/vk_hotload.rs
```diff
@@ -0,0 +1,246 @@
...
+        vert_shader: vs,
+        frag_shader: fs,
+        device,
+    };
+    update_pipeline(&mut model);
+    model
+}
+
+fn update(_app: &App, model: &mut Model, _: Update) {
+    let shader_msg = model.shade_watcher.rx.try_recv();
```
mitchmindtree: We only really care about the last `shader_msg` that was received, so maybe something like this would be better:
```rust
let shader_msg = model.shade_watcher.rx.try_iter().last();
```
This way we skip any backed up intermediary messages (e.g. in the case where multiple file system events occurred at once for some reason).

freesig:
I might add a few comments then merge this. Before doing that next pr there is a little more work to do on shade_runner's layouts. Basically there is a couple of types of descriptor sets that aren't implemented yet.

PR #300, opened 2019-05-05
travis rustfmt

### LINKED ISSUE #298: Setup travis to check `rustfmt` formatting
Now that #297 landed maybe it's best to enforce correct formatting in commits and have travis produce an error if formatting is incorrect?

One alternative would be to instead setup travis to add a commit that applies `rustfmt` in the case that a PR is incorrectly formatted when merged into master, but maybe this might cause some odd git conflict issues if someone branches from a previously merged branch without realising that travis has added a commit on top of the original branch?

cc @freesig

tpltnt:
Hi there,

this PR should close #298.

Cheers,
tpltnt

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,7 @@
 language: rust
+before_script:
+- rustup component add rustfmt
+
 addons:
     apt:
         packages:
@@ -7,7 +10,9 @@ rust:
     - nightly
     - beta
     - stable
+
 script:
+    - cargo fmt --all -- --check
     - cargo build -v
     - cargo test -v
     - cargo doc -v
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,7 @@
 language: rust
+
+cache: cargo
+
 before_script:
 - rustup component add rustfmt
 
--- a/examples/nature_of_code/chp_06_agents/6_01_seek.rs
+++ b/examples/nature_of_code/chp_06_agents/6_01_seek.rs
@@ -142,7 +142,7 @@ fn display(vehicle: &Vehicle, draw: &Draw) {
     } = vehicle;
     // Draw a triangle rotated in the direction of velocity
     // This calculation is wrong
-    let theta = (velocity.angle() + PI/2.0) * -1.0;
+    let theta = (velocity.angle() + PI / 2.0) * -1.0;
     let points = vec![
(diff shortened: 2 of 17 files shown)


mitchmindtree:
Hey @tpltnt thanks for the PR!

So to clarify, this will cause travis to fail if the PR would submit incorrectly formatted code?

Also I didn't know about the caching ability, that looks very useful! I wonder what travis uses for caching...

@freesig what are your thoughts on landing this in the meantime until we can work out how to get travis to automatically push formatting fixes to master?

tpltnt:
It should fail upon incorrectly formatted code. I made a demo-PR #301 to test it with travis.

tpltnt:
Just as a check-in: What is blocking this PR?

freesig:
I think we were after the ability for Travis to automatically run cargo fmt if the pr is formatted. This way people don't need to push changes that could be automated. I'm personally in favour of automating everything we can as the teams time is spread so thinly at the moment.

tpltnt:
I think I am missing something. My assumptions are that
1. "rustfmt" and "cargo fmt" are basically the same thing (for our purposes) and
2. travis can't make changes to code (e.g. by running "cargo fmt").
What do you mean exactly by "if the pr is formatted"? (My understanding is that rustfmt has been run before submitting the PR, which would make a check rather pointless)
What didn't I get?

freesig:
Oh I was thinking it could actually make another commit where it runs rust fmt (cargo fmt) if it hadn't been run. But perhaps this isn't possible?

tpltnt:
I see. The issue is that Travis would modify the repository and then push the changes. It is technically possible (if I read [this](https://stackoverflow.com/questions/26020494/automated-push-to-a-github-repo-with-travis) correctly). It would require an access token for the nannou-org/nannou repository, which i can't generate (since I am not one the the owners of this repo). This PR is the best thing I can do (without any special access). I would opt for merging it so we at least get the checks and also opening an issue so that people with the proper access can put the final touch on it.

freesig:
Ok sounds good to me, merging this now

mitchmindtree:
Just realised this might make it tricky to rebase v0.9 onto master again! I'll give it a go tonight.

mitchmindtree:
OK just finished rebasing, will force push to the v0.9 branch to update it once I finished running the tests locally.

PR #302, opened 2019-05-07
Shader Toy Style Example

freesig:
This is a example of how to pass data into a hot loaded shader. In the same fashion as shadertoy.
I still need to work out what the best way to send data is. __[These](http://vulkan-spec-chunked.ahcox.com/ch13s01.html)__ are our options plus push constants.


freesig:
Closing this because the rebase is too large and im just going to recreate the branch

PR #331, opened 2019-06-06
Separate audio API into a `nannou_audio` crate

mitchmindtree:
The motivation behind this is to avoid flooding nannou itself with APIs
for every kind of creative I/O we eventually implement. Nannou's build
times are already very unwieldy and I think this could easily get
out of hand if we continue to add everything into the one crate. This
change slightly improves build times for users who don't need or care
about audio.

In a similar manner, the laser API also exists in its own crate called
[lasy](https://github.com/nannou-org/lasy). Maybe the name should be
changed from `lasy` to `nannou_laser`, and we can make this a consistent
theme for supporting different higher-level creative I/O stream APIs
(e.g.  `nannou_dmx`, `nannou_midi`, etc)?

The `simple_audio.rs` example has been updated to demonstrate use of the
new `nannou_audio` crate.

cc @freesig @JoshuaBatty what are your thoughts on this? I've been
meaning to get around to this for a while. My original concern was that
we might not be able to keep an "as-simple" API if we were to move
everything into its own crate, but I'm pretty happy with how the
example's looking in this case. Pretty much everything is identical
except the audio API has been removed from the `app.audio` field to its
own type in its own crate.


freesig:
Yeh I think this is a good idea.
It will also help with having things under feature flags.
Like it could be cool to have everything but nannou itself under a feature but enabled by default (Or a large subset atleast).
That way users can pick exactly what they want to include for each project that they are working on.
This could help with compile times and binary sizes.

JoshuaBatty:
Nice the updated example still looks super simple enough. I'm also a fan of going with the ```nannou_midi``` type name space for all the crates we want to seperate from nannou's core.

mitchmindtree:
Sweeeeet!

Maybe as this is a pretty hefty breaking change I'll leave this up for a bit until we've announced 0.9 and done some more non-breaking stuff on 0.9 first.

mitchmindtree:
OK the 0.9 ship sailed successfully and there's a growing desire for quite a few changes that will involve some breakage on the horizon (palette crate update, addition of a loop mode, etc) so I think it's safe to start merging these in prep for a 0.10!

PR #340, opened 2019-06-11
Offline Example

MacTuitui:
One of the examples featured on the website, self-contained. A few warnings that I don't know how to get rid of, and some things that are hard-coded but should not. It's been a while since I did this. Comments welcome!


mitchmindtree:
Sorry about the delay on feedback @MacTuitui!

I think travis is complaining about incorrect formatting. Would you mind running `cargo fmt`?

mitchmindtree:
Thanks @MacTuitui!!

PR #353, opened 2019-06-19
Proposal for NTimes

CGUTA:
I duplicated the Wait loop mode to get to know the code and because I think it is close (in edits) to what the Ntimes loop mode must be.

I moved the update execution inside an if. It only runs if there are pending updates within the specified framerate. 

The window is still drawn and window events are still received, just no update is processed.

I am getting warnings because now `mut number_of_updates: usize,` is not used. But I could not manage to remove it, I do not understand what `ufe` is in `number_of_updates: ufe,`. 

In this way the method guarantees that only the specified update events will be called. Which can be called as necessary to respond to events.


mitchmindtree:
Thanks a lot for this @CGUTA! This is a great help :)

I made a PR at CGUTA/nannou#1 to tweak some things on this PR, if you can look over and merge those then I'll merge this once I've looked into this `panic!` I'm running into.

mitchmindtree:
@JoshuaBatty do you remember there being some nature of code examples or generative gestaltung examples that required a "draw once" kinda thing? Maybe it's worth trying to get those updated in this PR too so they don't lag behind?

JoshuaBatty:
I'll have a look now through the NOC examples and see which ones where waiting for this feature.

JoshuaBatty:
Hi @CGUTA I just made a PR to your branch that adds in 4 new recursion examples that take advantage of this new ```loop_once()``` mode.

mitchmindtree:
It looks like this needs a `cargo fmt` after merging that last PR, then should be ready to go :)

src/app.rs
```diff
@@ -1436,6 +1479,98 @@ where
     }
 }
 
+// Run the application loop under the `Wait` mode.
```
mitchmindtree: Oh woops, just realised this comment needs to be fixed.

PR #355, opened 2019-06-19
[WIP] Update palette crate to version 0.4.

### LINKED ISSUE #72: `color::named` consts are Srgb values but interpreted as linear RGB.
This should be fixed by updating to the latest palette crate.

### LINKED ISSUE #345: Colors in docs are not completely implemented
The docs mention a host of colors that are not actually available in the `nannou::color::named` namespace.
For instance, `nannou::color::named::ALICEBLUE` does not actually exist.
Should these still be implemented, or should the tango colours that exist in the source now not be expanded upon?

mitchmindtree:
The biggest update to account for here was the distinction between
`RgbSpace`s. Nannou now uses the `Srgb` standard encoding as a default
throughout the crate which should more accurately represents nannou's
provided set of named colors (#72) and likely close match users
expectations of the behaviour.

Closes #72.
Closes #345.

## TODO:

- [x] Decide whether or not to re-add the `Rgb` and `Rgba` types to the
  `color` module? Or should we encourage use of the `Srgb` and `Srgba`
  types instead for clarity? *Spoke with mactuitui and decided it would
  be nice to re-add them to reduce friction for new users*.
- [x] Add a suite of super simple constructor functions to the color
  module (e.g. `hsl`, `hsla`, `rgb`, `rgba`, etc) that produce color
  types that are easily compatible with nannou's `Draw` API.
- [x] Decide if we should keep the Tango based `named` colors provided
  by nannou (see [here][1]) or only use those provided [by palette][2].
  Alternatively we could keep both, but this would mean removing the
  glob palette import to fix #345 and re-exporting `palette`'s named
  colors into our own named module. *Decided to remove tango after
  discussion in slack - palette selection likely more standard and
  recognisable as they map to CSS colors*.
- [x] Decide what the most convenient type is for `named` colors. We
  should probably change them to the same type as `palette`'s named
  colors, but also ensure that these can be conveniently used within all
  areas of the `Draw` API. The problem with using `Srgb<u8>`
  is that it's not currently trivially compatible with the `Draw` API, as the
  `Draw` API requires colors with floating point components. Maybe
  there's a way to avoid imposing this restriction and allow for byte
  representations... *OK managed to allow for colors with byte
  component representations and have confirmed the palette named
  colors are now easily compatible with the draw API*.
- [x] Run all examples using `run_all_examples` test.
- [x] Clarify whether we should be useing `Srgb` or `Linear<Srgb>` RGB
  encoding for nannou's default draw API. I've opened up
  ogeon/palette#133 to help clarify my understanding on this.

## Future PR

- [x] Create a `colors.rs` example that demonstrates all named colors
  within the crate. Ensure that these colors visibly match their
  references.
- [x] Update examples to use shorthand constructors where possible.

cc @SjorsVanGelderen.

[1]: https://github.com/nannou-org/nannou/blob/master/src/color.rs#L12
[2]: https://docs.rs/palette/0.4.1/palette/named/index.html


freesig:
[Heres](https://matt77hias.github.io/blog/2018/07/01/linear-gamma-and-sRGB-color-spaces.html) a explanation of colour space as related to rendering.
Please correct me but I think sRGB is a colour space where the colours appear perceptually linear but are not mathematically linear.
Looks like the pipeline will make these conversions to linear when needed.
I think the only time we need to be careful is when sampling or copying to a buffer.
We just need to make sure that when sampling from an image or copying from a buffer to an image that the sample format is on the image that palette is using.

mitchmindtree:
Just found this excellent discussion on linear and non-linear sRGB. This along with the responses at Ogeon/palette#133 help to clear up the main confusions I had. The main gist I'm getting is that linearized sRGB is better suited for operations, while non-linearized sRGB is better suited to storage. 

*It's important to note that both are still sRGB. As the CG post I linked above discusses, it doesn't make sense to talk about "linear RGB" without mentioning the encoding (which in this case is sRGB), as knowing the RGB encoding is necessary to know the colour space intrinsics - Primaries, Whitepoint and Colour Component Transfer Functions.*

So the pipeline should probably look something like this:

1. **Non-linear sRGB** from colour storage sources like hex value constants or image files.
2. Into **Linear sRGB** for operations, blending, processing.
3. Into **Non-linear sRGB** ready for swapchain presentation. It's possible this could be different on certain systems but non-linear sRGB seems to be the most consistently supported default.

Taking this into consideration, I think I should:

- [ ] **Change the target colour for the Draw API to be linearized sRGB** so that it is better prepared for the operations that are likely to follow, including blending, processing in the fragment shader, etc. Most colour conversions when converting *to* sRGB seem to require first passing through linearized sRGB before they can be converted to non-linearized anyway, so this may improve CPU efficiency slightly, particularly when receiving HSL or HSV colour inputs.
- [ ] **Double check that the colour formats specified within the render pass attachment descriptions in the draw `backend::vulkano` module match the colour format of the vertex input (which should be Linear sRGB)**. Currently, we use the swapchain's colour format for the attachment image colour formats, which is non-linear sRGB. This could potentially be a significant problem when operating on colours in the fragment shader, as mathematical operations are better suited to the linearized format. I'm not certain that the colour format doesn't get automagically converted from the fragment shader output to the colour format specified for the attachment image, but after a quick look through the vulkano docs I can't find anywhere that lets the user specify the colour format of the vertex data, so I'd assume that the colour output of the fragment shader must match the colour format of the target renderpass attachment image? Need to investigate.
- [ ] **Check that we don't have to do the same as above for the UI.** Seeing as we're using conrod, the UI has its own colour thing going on entirely. It expects non-linear sRGB when colouring widgets and also produces non-linear sRGB colours in the vertices of the tesselated output. I think this is OK for now, as there are no operations applied to any of the UI colors in the graphics pipeline. As a result, it should be fine to "pass them through" directly to the swapchain image. This will change though when support for blending gets added to the Ui, but we're probably better off addressing this as a part of the GUI editor project (#2).
- [ ] **Ensure that the image attachment for drawing to the swapchain image still uses the swapchain color format** and that vulkan automatically takes care of the conversion from linear sRGB to non-linear sRGB during the renderpass that writes the intermediary image to the swapchain image.
- [ ] **Check if any examples need to be updated**. If any of the vk examples happen to be applying operations to colour in the fragment shader, we should probably ensure that the input colours are linearized sRGB and that the target attachment image in the render pass expects this format.

 cc @JoshuaBatty @freesig remember when we were doing the `hsl_to_rgb` fragment shader color conversions and seeing a strange break in the hue values at the crossover between 2PI and 0 radians in the mirrors and tunnel projects? I wonder if that was because we were feeding non-linear sRGB vertex data into the graphics pipeline, when the operations in that function were designed to work on linear sRGB data...

mitchmindtree:
OK! That turned out to be a seriously epic overhaul, but I think it's good to go!

Here are the highlights taken from the CHANGELOG update:

- Update from palette 0.2 to 0.4.
- Add shorthand color constructors to the `color` module (e.g. `rgb`, `hsla`, etc) and expose them from the prelude.
- Remove nannou named colors in favour of using palette's `named` prelude which match the CSS colors.
- Added a `named_color_reference.rs` example for finding suitable colors and to
  act as a test of color accuracy. I've been able to use this to confirm that colours are at least consistent across the UI and Draw APIs on both my linux laptop and josh's macbook pro and match the CSS spec.
- Change the `Frame` image type from the swapchain color format (non-linear
  sRGB) to a linear sRGB format for better consistency across platforms.
- Add a `Window::rect` method.

mitchmindtree:
Just got a hot-tip [from a /r/vulkan user](https://www.reddit.com/r/vulkan/comments/cfl7zs/what_is_the_most_efficient_way_to_write_a_linear/euc40s9/) to use the BlitImage command for writing the linear sRGBA frame image to the swapchain image as it applies the lienar->non-linear conversion after filtering, and it seems to work nicely!

PR #358, opened 2019-06-30
Refactor Nature of Code's 2_5 example

pepyakin:
There is some duplication in code I found.

The new code is not identical to the previous version since `m` was picked randomly from `1.0..4.0`, but `new_random` picks it from `0.5..4.0`


mitchmindtree:
Perfect, thanks @pepyakin!

mitchmindtree:
Oh actually, would you mind running `cargo fmt`? I think that's what travis is complaining about :)

mitchmindtree:
Thanks!

PR #359, opened 2019-07-02
Add an example of using audrey for audio file playback

mitchmindtree:
This new example uses audrey to play back a WAV file of a thumb piano. A new instance of the sound begins playing every time spacebar is pressed, demonstrating very basic polyphonic playback.

It should also be possible to play FLAC, Ogg and ALAC using audrey - audrey will try to work out what the file format is from the file extension.

Find out more about audrey at [the repo](https://github.com/RustAudio/audrey) or on [docs.rs](https://docs.rs/audrey).


mitchmindtree:
I should also add this example assumes the audio file has the same sample rate as the audio stream itself. To make sure this is correct, you can check the sample rate of both the stream and file and use the [sample](https://github.com/RustAudio/sample) crate to interpolate between the two rates (I think there's an example of this at the repo).

Eventually though, hopefully we can just make some higher-level APIs so that users don't have to worry about this!

PR #361, opened 2019-07-04
Update rand to version 0.7

mitchmindtree:
Should check if this will create conflicts with the `noise` crate import before merging.


mitchmindtree:
It looks like the `noise` crate is already using a `rand` that is a few versions old. If this causes issues we can take a look at contributing an update to the `noise` crate.

Issue #365, opened 2019-07-17
example broken


edap:
Hey! i wanted to try out nannou but the example that appears in https://guide.nannou.cc/getting_started/running_examples.html looks broken:
When I run `cargo run --release --example simple_draw` I receive:

```
   Compiling shade_runner v0.1.1
   Compiling conrod_vulkano v0.65.0
error[E0658]: multiple patterns in `if let` and `while let` are unstable (see issue #48215)
   --> /Users/da1/.cargo/registry/src/github.com-1ecc6299db9ec823/shade_runner-0.1.1/src/watch.rs:195:9
    |
195 | /         if let Ok(notify::DebouncedEvent::Create(_)) | Ok(notify::DebouncedEvent::Write(_)) =
196 | |             notify_rx.recv_timeout(Duration::from_secs(1))
197 | |         {
198 | |             loader.reload();
199 | |         }
    | |_________^

error: aborting due to previous error

For more information about this error, try `rustc --explain E0658`.
error: Could not compile `shade_runner`.
warning: build failed, waiting for other jobs to finish...
error: build failed
```

freesig:
That's a bit weird, looks like shade_runner isn't compiling. Kind of weird because I haven't changed anything in it but I'll check it out

freesig:
Which version of rust are you using? It might be worth running `rustup update`

mitchmindtree:
Hi Davide! Super exciting to have you visit :)

As @freesig says it does look a little like you might have an old version of Rust installed, is that possible? Nannou is pretty bleeding edge at the moment so there's a good chance updating Rust might fix this! Feel free to [join us on slack](https://communityinviter.com/apps/nannou/nannou-slack) too if you feel like chatting in realtime :)

edap:
Hello Mitch, hello Tom, thanks for your support ;). Yes, `rustup update` fixes the first error. Now I have another one, that pops up after the moltenVK installation (i am on mac)

```
Unpacking file into ~/.vulkan_sdk
The Vulkan SDK was successfully installed at /Users/da1/.vulkan_sdk
Setting environment variables
bash: line 2: /Users/da1/.bash_login: No such file or directory
Installation complete :D
To update simply remove the '~/.vulkan_sdk' directory
[mvk-error] VK_ERROR_FORMAT_NOT_SUPPORTED: VkFormat VK_FORMAT_D16_UNORM is not supported on this platform. Using VkFormat VK_FORMAT_D32_SFLOAT instead.
[...]

PR #367, opened 2019-07-20
Remove depth format constants in favour of querying supported formats

### LINKED ISSUE #303: DEPTH_FORMAT should not be const D16Unorm
[This](https://github.com/nannou-org/nannou/blob/27464de0c359b3fcfc3c132343f468f6ac6296af/src/app.rs#L46) shouldn't be a const.
It needs to be verified that the host system actually supports the value. It should probably be selected from a list of supported formats from the host system.

mitchmindtree:
The current implementation is a bit of a hack but at least enables users
who are blocked by this while we work out a proper solution in the
meantime!

Closes #303.
Related to #365 (should hopefully close it).

cc @freesig does look alright for now?

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -33,12 +33,6 @@ use winit;
 #[cfg(all(target_os = "macos", not(test)))]
 use moltenvk_deps as mvkd;
 
-// TODO: This value is just copied from an example, need to:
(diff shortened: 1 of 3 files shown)


freesig:
Looks good to me but I'll give it a test and report back. Sorry about the delay in responses

Issue #366, opened 2019-07-18
[IDEAS] list on features & thoughts for nannou from a beginner perspective


lewislepton:
howdy folks,

well im looking at nannou & have been for the past few weeks. really like it. though im no rust person [but have been interested in using it for a while]. just need to get used too it more than anything.

but I do have questions & thoughts on nannou. this is purely coming from a beginner of it & also as someone who has been doing coding with openFrameworks for a while. also [& more importantly] from a realm of making things understandable to new users that may come across this platform. how it can be made easier to understand etc.

so some of these questions may seem pedestrian at first & can easily be looked down upon by others for the simplicity of them, or even the misunderstanding of how something works. but its purely to make things not only easier to understand but maybe to open minds in ways that could help the platform become more user friendly for utter beginners & people shopping around for a new coding playground to delve into ;)

lewislepton:
one thing that pops out at me when creating a project is the rendering of it. making the project is fine to a degree after looking at what things do. I know there has been talk of a project generator, which would be fantastic. so im not touching on project starting.

but one catch I think is a bit head scratching & wondering if anything could be done to lessen it or make it part of the backbone code is how it renders, or gets the drawn frame on screen.
with `frame` right at the end of the view function. this is a weird little catch & im trying to think of ways that this little sucker could be put into the background of nannous  code.

```
fn view(app: &App, model: &Model, frame: Frame)->Frame {
  let draw = app.draw();
  let time = app.time;
  draw.background().color(Rgb::new(0.13, 0.13, 0.13));
  draw.ellipse().x_y(app.mouse.x * time.cos(), app.mouse.y * time.sin());
  draw.to_frame(app, &frame).unwrap();
  frame ///<---- this wee mofo right here
}
```

because whilst 'yes' it is to show the drawn frame. but it feels & looks like we are already doing it with `draw_to_frame(app, &frame).unwrap();`. maybe even doing that twice if we bring in UI to draw to the frame.
so it feels as if we are saying it twice, just to make sure it's listened & clean its room.

in some ways this feels more understandable in my head on if frame is still in
```
frame.begin();
/// all the other code that has circles & cack
frame.end();
```

but rather than going the path of having a `begin()` or `end()`, I was thinking if its possible to even have `frame` in the background of tho source code of nannou?
if not, that is fine. but its one catch that really baffled me [maybe others] & did not stick out until I sat back, then saw that little sucker winking at me.

mitchmindtree:
[...]

PR #368, opened 2019-07-20
Change `view` function signature to take `Frame` by reference rather than value

### LINKED ISSUE #328: Update docs to not mention OpenGL
[window](https://docs.rs/nannou/0.9.0/nannou/window/index.html) still says `OpenGL` window

mitchmindtree:
This was suggested by @lewislepton in #366 and I think it makes a lot of
sense in terms of API consistency and keeping things simple for new users.
See #366 for the lead-up discussion, history and reasoning behind the change.

Also closes #328.

cc @freesig and @JoshuaBatty what are your thoughts on this?

### DIFF
--- a/examples/all_functions.rs
+++ b/examples/all_functions.rs
@@ -54,9 +54,8 @@ fn event(_app: &App, _model: &mut Model, event: Event) {
 
 fn update(_app: &App, _model: &mut Model, _update: Update) {}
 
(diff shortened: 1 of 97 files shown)


freesig:
Looks good to me. I always thought the reason it was by value was because you were mutating it but I guess that's happening in some kind of RefCell anway?
I'm always in favor of simplifying the api in was that mean there's less for the user to have to remember

Issue #194, opened 2018-09-17
Allow easier translation/rotation of groups of primitives


SuperCuber:
# Problem
There's no easy way to translate+rotate a group of primitives that represent a single object.

In ProcessingJS for example you would be able to translate and rotate the coordinate system itself, then draw the primitives, then restore the state to what it was before. Right now you need to calculate the position+rotation of each primitive yourself.

# Proposed solutions:
- Copy ProcessingJS's approach and allow manipulating the coordinate system (+ adding a builtin ability to push and pop the current state of the coordinate system to a stack)
- Create some `Canvas` primitive that can be drawn to, and then drawn using a position+rotation onto the real screen

---

I am not completely familiar with this project so if I missed the intended way to do it, then I'd love to be pointed to docs of it. (Couldn't find any example of such a thing)

mitchmindtree:
Thanks for opening this!

Totally agreed there needs to be an easier solution for rotating multiple primitives at once!

While the processing approach of using a global rotation is very easy to add in and is easier to use with small sketches, it can be tricky to keep track of as programs grow and it can get easy to unexpectedly break things by adding or removing rotations or refactoring.

Maybe we can achieve something similar that feels as easy to use by adding in some explicit translation/rotation types (matrices under the hood) that can be chained together and used to position primitives as an alternative to the positioning and orientation methods.

SuperCuber:
@mitchmindtree I haven't personally encountered the problem of losing track since most usage of that feature goes along the lines of:


```
push_current_coordinate_system()
rotate(...)
scale(...)
translate(...)

draw_primitives()
draw_primitives()
draw_primitives() // Relative to 0,0

pop_coordinate_system()
```

Note that that code above doesn't need to start from an "identity matrix" state, it can be put in a function and used to draw sub-objects of objects for example. I think it would be useful to enforce/automate this pushing/popping using a constructor and destructor though.

freesig:
I agree that that global translates are easier with simple projects but they can get hard to track as the project grows. Especially with multiple threads. 
It would be ideal if we could find a solution almost as easy to use but without the global state. We want nannou to be great for tiny sketches as well as large projects

SuperCuber:
@freesig Have you considered the Canvas solution? What do you think would be the pros/cons of that? To elaborate, Canvas can mimic the API of Draw and also have some sort of `.canvas(...)` function that allows drawing a canvas onto another canvas (Draw can be a Canvas too?).

So you would do something along the lines of
[...]

Issue #86, opened 2018-03-27
Port the Nature of Code examples over to run in Nannou


JoshuaBatty:
In order to both test the capabilities of nannou and it's API, and launch with some ridiculously handy tricks for users to reference, let's port of the Nature of Code examples written by Daniel Shiffman in Processing. 

Below are a list of examples by chapter that i'll tick off and leave notes as I go. 

**Chapter 1: Vectors**
- [ ] NOC_1_1_bouncingball_novectors
- [ ] NOC_1_2_bouncingball_vectors
- [ ] NOC_1_2_bouncingball_vectors_object
- [ ] NOC_1_3_vector_subtraction
- [ ] NOC_1_4_vector_multiplication
- [ ] NOC_1_5_vector_magnitude
- [ ] NOC_1_6_vector_normalize
- [ ] NOC_1_7_motion101
- [ ] NOC_1_8_motion101_acceleration
- [ ] NOC_1_9_motion101_acceleration
- [ ] NOC_1_10_motion101_acceleration
- [ ] NOC_1_11_motion101_acceleration_array

**Chapter 2: Forces**
- [ ] NOC_2_1_forces
- [ ] NOC_2_2_forces_many
- [ ] NOC_2_3_forces_many_realgravity
- [ ] NOC_2_4_forces_friction
- [ ] NOC_2_4_forces_nofriction
- [ ] NOC_2_5_fluidresistance
- [ ] NOC_2_5_fluidresistance_sequence
- [ ] NOC_2_6_attraction
- [ ] NOC_2_7_attraction_many
- [ ] NOC_2_8_mutual_attraction
- [ ] NOC_02forces_many_attraction_3D
- [ ] NOC_02forces_many_mutual_boundaries
- [ ] Exercise_2_10_attractrepel
- [ ] Extra_instantforce

**Chapter 3: Oscillation**
- [ ] NOC_3_01_angular_motion
- [ ] NOC_3_02_forces_angular_motion
- [ ] NOC_3_03_pointing_velocity
- [ ] NOC_03spring_exercise_sine
- [ ] NOC_3_04_PolarToCartesian
- [ ] NOC_3_04_PolarToCartesian_trail
- [ ] NOC_3_05_simple_harmonic_motion
- [ ] NOC_3_06_simple_harmonic_motion
- [ ] NOC_3_07_oscillating_objects
- [ ] NOC_3_08_static_wave_lines
- [ ] NOC_3_09_exercise_additive_wave
[...]

PR #375, opened 2019-08-03
lyon 2D tessellation integration - Fixes many 2D bugs, adds support for stroke tessellation, Paths, curves and more.

### LINKED ISSUE #167: Add an `.outline()` method to some `Draw` types using polyline geometry.
Outlines are commonly used throughout Processing and as a result, this is blocking quite a few of the examples #86.

This could be done by creating a polyline to surround the geometry using the `geom::line::join::miter` items.

Shapes that should support outline include:

- [ ] `rect`
- [ ] `ellipse`
- [ ] `polygon`
- [ ] `quad`
- [ ] `tri`

Styling options for the outline should include:

- `thickness`
- `color`
- `join`

### LINKED ISSUE #185: Polyline bugs - strange behaviour at extreme join angles, freezes on `1` point
While implementing a NOC that requires a line I noticed some issues with Polyline.
1. There's no warning for if you build a Polyline with a single vertex. It just freezes. This would seem like an avoidable mistake but if you are building with loops it's easy enough to make.
I'd like to put a panic! on < 2 vertices with a help message if that's ok with everyone?
2. There seems to be big "bleeds" where it looks like it's trying to fill a sharp angle. I might be trying to use this wrong? My goal is a series of vertices linked together by a line
image
```Rust
    let thickness = 1.0;
    if history.len() > 1 {
        let vertices = history.iter().map(|v| {
            geom::vertex::Rgba(pt2(v.x, v.y), Rgba::new(0.0, 0.0, 0.0, 1.0))
        }); 
    draw.polyline().vertices(thickness, vertices);
```

### LINKED ISSUE #245: Polygon and convexity?
I'm trying to draw a polygon with the following code:

```rust
let mut points = Vec::new();
let s = 5;
for k in 0..s {
    let frac = (k as f32)/(s as f32);
    let angle = frac*TAU;
    let r = (frac-0.2).abs() * 300.0;
    points.push(pt2(r*angle.cos(), r*angle.sin()));
}
draw.polygon()
    .points(points)
    .rgba(1.0, 1.0, 1.0, 0.4);
```

screen shot 2019-01-21 at 14 31 08

But it looks like the order of the vertices is changed somehow. I'm lost as to why (convexity? triangulation?).

### LINKED ISSUE #246: Allow for concave polygons via `draw.polygon()`
See #245 for motivation.

### LINKED ISSUE #253: Investigate using the `lyon` crate under the hood for handling 2D geometry
I've been discussing a lot of the shortcomings of nannou's current WIP 2D implementation with @MacTuitui and remembered that I've been meaning to look into [lyon](https://crates.io/crates/lyon) as a solution for a while now.

These are all the related issues that we could potentially solve or make progress on:
#164 
#167 
#185 
#245 
#246 

Things to investigate:

- How to implement a lyon tesselator using the vulkano backend. Ensure we can fallback to CPU tesselation if GPU does not support it.
- Work out what the best level is for integration. E.g. can we replace the 2D geom modules entirely? How much work is there in allowing existing nannou point, vector, etc types to interop with the lyon API?

### LINKED ISSUE #281: Polyline flashing with weird shapes
I am trying to learn nannou, and I attempted to make a line that follows the y position of your mouse.

The line flickers as seen in the gif below:

Flicker

I came up with the following source code:

```rust
extern crate nannou;

use nannou::prelude::*;

fn main() {
    nannou::run(model, event, view);
}

struct Model {
    y_positions: Vec<f32>,
}

fn model(app: &App) -> Model {
    // Start in `Wait` mode. In other words, don't keep looping, just wait for events.
    app.set_loop_mode(LoopMode::rate_fps(60.0));
    // Set a window title.
    app.main_window().set_title("`LoopMode` Demonstration");
    Model {
        y_positions: Vec::new(),
    }
}

fn event(app: &App, mut model: Model, event: Event) -> Model {
    match event {
        Event::Update(update) => {
            model.y_positions.push(app.mouse.y);
            if model.y_positions.len() > 300 / 5 {
                model.y_positions.remove(0);
            }
        }
        _ => (),
    }
    model
}

// Draw the state of your `Model` into the given `Frame` here.
fn view(app: &App, model: &Model, frame: Frame) -> Frame {
    let window = app.window_rect();
    let draw = app.draw();

    draw.background().color(BLUE);

    if model.y_positions.len() > 1 {
        let half_thickness = 0.5;
        let vertices = (0..model.y_positions.len())
            .map(|i| pt2(((i as f32) * 5.0) - window.w() / 2.0, model.y_positions[i]))
            .enumerate()
            .map(|(i, p)| {
                let r = 1.0;
                let g = 0.0;
                let b = 0.0;
                let rgba = nannou::color::Rgba::new(r, g, b, 1.0);
                geom::vertex::Rgba(p, rgba)
            });

        draw.polyline().vertices(half_thickness, vertices);
    }

    frame
}
```

Is this something I am doing wrong, or a bug with nannou?

### LINKED ISSUE #335: Polyline with caps_round()
Polyline with caps_round() is this possible?
And configuring joints like bevel, miter, joint?
Cheers for the beautiful project!

### LINKED ISSUE #371: Switch to `lyon` for 2D geometry tesselation
This is the actionable issue follow-up to #253. I'll update this comment with new steps as I work out the seemingly best way to approach it.

It's not yet clear to me whether or not we should remove nannou's 2D `geom` modules entirely in favour of lyon. While lyon is far better for tesselation, some of nannou's 2D geom types like `Rect`, `Point`, `Range` etc provide a whole suite of useful methods for easy integration with the rest of nannou's API and to be a little friendlier to new users.

For this reason, it might make sense to first replace the tesselation (triangulation) implementations first.

- [x] lyon Point and nannou Point interop conversions.
- [x] Create a simplified, nannou-friendly abstraction around `lyon::path`.
- [x] Add implementation of [`GeometryBuilder`](https://docs.rs/lyon_tessellation/0.14.0/lyon_tessellation/geometry_builder/index.html#generating-a-completely-custom-output) for building a nannou mesh.
- ~~Consider API options for providing fill and stroke options that makes the most of `mesh` and `path` but also allows specifying positioning, orienting, etc methods for basic shapes. Worth considering the draw "context" idea mentioned in #194.~~ Going to save this for a future PR.
- ~~Consider replacing `IntoDrawn` with `GeometryBuilder`-like API. This might address current issue with incorrect index ranges when drawing primitives that use the intermediary mesh (meshs, paths).~~ Saving for a future PR.
- [x] Add `draw.path()` API wrapping lyon path with curve support, etc.
- Replace `draw` primitive tesselation with lyon tesselation:
    - [x] polyline (replaced with `draw.path()`).
    - [x] polygon (working with single colour, coloured vertices blocked by nical/lyon#488).
    - [x] ellipse
    - [x] line
    - [x] quad
    - [x] rect
    - [x] tri

mitchmindtree:
This overhauls the `Draw` APIs tessellation approach to take advantage of the `lyon` crate.

Highlights of the PR include:

- A new `geom::path()` API has been added that allows for building 2D vector graphics paths as an iterator yielding `lyon::path::PathEvent`s. This adds support for curves, arcs, sub-paths and more.
- A `draw.path()` API has been added to allow for taking advantage of paths via the `Draw` API. `draw.path().stroke()` produces a path that will be rendered via stroke tessellation, `draw.path().fill()` produces a path that will be rendered via fill tessellation.
- The `draw.polyline()` and `draw.line()` APIs are now implemented in terms of `draw.path().stroke()`.
- All known polyline bugs should be fixed. Closes #185. Closes #281. Closes #335.
- `draw.polygon()` has been updated to use lyon's `FillTessellator` allowing for concave shapes. Closes #245. Closes #246. 
- `draw.polygon()` now supports optional stroke tessellation of its outline and includes a suite of stroke option builder methods including line join types, stroke weight, stroke color, etc. See the `SetStroke` method docs to find all new methods now available.
- `.no_fill()` and `.stroke(color)` can be called on all polygon types to indicate that no fill tessellation is required or to specify stroke color respectively.
- All other `draw` API polygons (`rect`, `quad`, `tri`, `ellipse`) have been implemented in terms of `draw.polygon()`, allowing them to take advantage of the same stroke tessellation options. Closes #167. 
- The line `thickness` methods have been replaced with `stroke_weight` and `weight` methods.
- Fixes a pretty severe bug where any draw primitives that use the intermediary mesh would produce incorrect triangulation indices if they weren't the first instance to be created.

Regressions as a result of this PR include:

- `draw.polygon()` will temporarily lose support for individually colored vertices. This is due to limitations with lyon's `FillTessellator`, however these are in the process of being addressed. See nical/lyon#488.
- `draw.tri()` and `draw.quad()` now expect `Point2`s instead of `Point3`s. This was a trade-off in order to take advantage of the lyon tessellators which only support 2D geometry. Currently, the draw API's 3D story is very limited anyway, and this can likely be revisited as a part of a larger 3D tessellation overhaul. For now, `draw.mesh()` can still be used for drawing arbitrary 3D via the `draw` API.

As a side note, these new stroke tessellation features should open up the ability to implement a lot of the remaining nature of code examples! #86.

Closes #253.
Closes #371.

The next focus for the Draw API should likely be #194, but that can be addressed in a future PR.


mitchmindtree:
Looks like travis is just failing due to rustfmt missing from the latest nightly toolchain but will be fixed in the next rust-lang/rust#62805.

PR #388, opened 2019-08-18
 Add a `text` module providing various items useful for text layout. Begin work on `draw.text("some text")` API (WIP).

### LINKED ISSUE #77: Add `text` logic module and support for drawing text via the `Draw` api
`draw.text("asdfasdfafd")`

mitchmindtree:
The new text module extends the `rusttype` crate with simplified interop
with nannou's geometry types and provides functionality for handling
layout of aligned and justified multi-line text.

The new `draw.text("some text")` API takes advantage of this new
module to allow for displaying multi-line text, handling line
wrapping, alignment, justification, font sizes and different fonts.

TODO:

- [x] Add a simple text layout builder type and high-level API to text mod.
- [x] Add support for converting text into lyon path events to allow for
path tessellation.
- [x] Add a text buffer to `draw::State` for buffering text. This allows
to avoid the need for specifying complicated lifetimes in the `Text`
type itself in a fairly efficient manner.
- [x] Add a glyph cache to the `draw::backend::vulkano::Renderer`. Can
likely copy the implementation of the Ui's glyph cache.
- [x] Create a `positioned_glyphs` iterator that yields all glyphs for a
single text instance.
- ~~Add a font map to the `Draw` API state.~~ *Bailing on this in favour of using fonts directly.*
- [ ] Write `positioned_glyphs` into mesh while updating the CPU half
of the glyph cache. For now, assume tex_coords channel in main draw
mesh refers to text cache image sampler coordinates.
- [x] Add a `text` example demonstrating alignment, justification, etc.

Closes #77.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,7 @@ notosans = { version = "0.1", optional = true }
 palette = "0.4"
 pennereq = "0.3"
 rand = "0.7"
+rusttype = "0.7"
 serde = "1"
 serde_derive = "1"
 serde_json = "1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -87,6 +87,9 @@ path = "examples/simple_polygon.rs"
 name = "simple_polyline"
 path = "examples/simple_polyline.rs"
 [[example]]
+name = "simple_text_path"
+path = "examples/simple_text_path.rs"
+[[example]]
 name = "simple_ui"
 path = "examples/simple_ui.rs"
 [[example]]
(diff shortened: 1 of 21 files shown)


mitchmindtree:
OK, this should be ready to go for now!

I'm leaving the GPU glyph caching out for now as I need to give some more thought to how to best go about it. The majority of the caching code is ready, but the tricky part is working out how to fit it nicely into the Draw API. As a result, `draw.text("foo")` will be very slow at first. We'll address this in a future release, possibly even in a patch version to this upcoming release.

Issue #392, opened 2019-09-04
Invalid mouse position when switching window


mft-ableton:
When at least two windows are being used, when you click inside another window (without moving the mouse after clicking), `app.mouse.window` and `app.mouse.buttons` (pressed state)  are updated, but not `app.mouse.x` nor `app.mouse.y`. 
So these two values are invalid, as well as the position in the pressed state of the mouse (which uses the position of `app.mouse`).
This behaviour can be observed looking at the output of the multi window example `multi_window.rs`.

mitchmindtree:
Would you mind also sharing which platform you're on? This might help to work out the source of the issue in the necessary platform on winit, and will allow us to check if this is the case on other platforms as well.

mft-ableton:
Sure: macOS Mojave, Version 10.14.6

JoshuaBatty:
This seems to be addressed now with the PR #452 to wgpu. Now experiencing some slow frame issues but i'll open another issue for that.

PR #393, opened 2019-09-04
Bugfix: mouse position validity

mft-ableton:
fixes https://github.com/nannou-org/nannou/issues/392

When at least two windows are being used, when you click inside another window (without moving the mouse after clicking), app.mouse.window and app.mouse.buttons (pressed state) are updated, but not app.mouse.x nor app.mouse.y.
So these two values are invalid, as well as the position in the pressed state of the mouse (which uses the position of app.mouse).
To deal with that, we define the bool `app.mouse.is_position_valid` which assesses the validity of the position.

NB: ideally the winit should send a CursorMoved event before any pressed event (I saw in the git history of winit that they did that at some point of their history, but somehow it is not the case here)... maybe it is a question of configuring winit correctly?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ target/
 **/*.rs.bk
 Cargo.lock
 .DS_Store
+.vscode/
--- a/src/app.rs
+++ b/src/app.rs
@@ -1873,9 +1873,13 @@ where
                     app.mouse.x = x;
                     app.mouse.y = y;
                     app.mouse.window = Some(window_id);
+                    app.mouse.is_position_valid = true;
                 }
 
                 winit::WindowEvent::MouseInput { state, button, .. } => {
+                    if app.mouse.window != Some(window_id) {
+                        app.mouse.is_position_valid = false;
+                    }
                     match state {
                         event::ElementState::Pressed => {
                             let p = app.mouse.position();
--- a/src/state.rs
+++ b/src/state.rs
@@ -89,6 +89,9 @@ pub mod mouse {
         pub x: S,
         /// *y* position relative to the middle of `window`.
         pub y: S,
+        /// In special cases, the mouse position might be invalid
+        /// (for example: first press in unfocused window when using several windows)
+        pub is_position_valid: bool,
         /// A map describing the state of each mouse button.
         pub buttons: ButtonMap,
     }
@@ -125,6 +128,7 @@ pub mod mouse {
             Mouse {
                 window: None,
                 buttons: ButtonMap::new(),
+                is_position_valid: false,
                 x: S::zero(),
                 y: S::zero(),
             }


mitchmindtree:
Thanks for the PR @mft-ableton! I'm in my home country with family at the moment and haven't had a lot of time to keep up with github, sorry about the delay!

> ideally the winit should send a CursorMoved event before any pressed event

Yeah I agree - I wouldn't be surprised if this has been addressed in the "EventLoop 2.0" overhaul that is currently undergoing at the winit repo. Once a stable version of winit 0.20 is released we'll update nannou and check if this is still an issue then.

In the meantime, I'm torn on whether we should add this `bool` or whether we should let the user track the mouse position however they like using the `mouse_moved` and `mouse_pressed` functions?

Alternatively, maybe rather than setting a `bool`, we can determine the new mouse position relative to the new window by using the `window.position()` methods on each window and the dimensions of each window?

mft-ableton:
All good :)

Yes, sure, I can track it on my own using `mouse_moved` and `mouse_pressed`. But then if we say it is up the user to track the cursor, what is the purpose of `app.mouse`? 
It is great for a single window, but in case of several windows I have to implement the `mouse_moved`, `mouse_pressed` for each window, so that I can know when a window is out of focus (since the last `mouse_moved` information for a window is outdated if another window has been focused in-between).
I think a less verbose alternative would be to reimplement the equivalent of `app.mouse` in my own model by listening the `Event::WindowEvent` events from the app (the code should be shorter since it allows to directly write logic based on the window id).

Can you tell me more about how to use `window.position()` to determine the new mouse position?
This would require to get a valid mouse position from somewhere in some coordinate system?

mitchmindtree:
> But then if we say it is up the user to track the cursor, what is the purpose of app.mouse?

Ahh I only meant to suggest that if the user is running into the same problem you are currently running into, that they can work around it in the meantime by using the `mouse_moved` and `mouse_pressed` functions, but only until we come up with a proper solution. I think this might be better than us adding this validity check `bool`, only to remove it later on once we fix the behaviour. I hope this makes sense!

> Can you tell me more about how to use window.position() to determine the new mouse position?

By subtracting one window's position from another, we can determine the offset required to update the `app.mouse.{x, y}` fields so that they are relative to the newly focused window. The real solution is not *quite* this simple though, as we also have to account for 1. DPI and 2. the fact that in nannou, mouse positions are relative to the centre of the window, not the top-left as is given to us by the underlying winit API (though this just means we need to account for the dimensions of the windows too). I'm a little busy to go into more detail just now but hopefully this gives an idea of the workaround I have in mind! If you don't beat me to it, I will hopefully get a chance to tackle this early next month, but feel free to have a go if you're interested.

mft-ableton:
> I think this might be better than us adding this validity check bool, only to remove it later on once we fix the behaviour. 

This makes totally sense 👍 

> If you don't beat me to it, I will hopefully get a chance to tackle this early next month, but feel free to have a go if you're interested.

So I'll close this PR. Thank you for your answers!
I don't think I will keep investigating for now, I guess I prefer experimenting further with Nannou, since it I just started playing with it few weeks ago.

Issue #374, opened 2019-08-02
Considering rendy and wgpu for use as nannou's graphics backend


freesig:
## Goal
Find out what is the current state of [Rendy](https://github.com/amethyst/rendy) to see if it's ready for use to use as a back end. If not establish how much work is remaining.
## Todo
- [x] Run some [examples](https://github.com/amethyst/rendy/tree/master/rendy/examples) and see how hard this is to do. 
- [x] Make a list of requirements we have currently using vulkano and see if we can match them in rendy. (nannou's vk examples serve as a good list).
- [x] Try and port some of the vulkan examples to rendy. This should help use get an idea of what is missing.
- [x] Talk to the people at Amethyst and establish some relationships so we can work together on implementing any remaining features.
- [x] Decide if we should go ahead with this project based on how much work is left vs vulkano.

mitchmindtree:
Someone inquiring about rendy learning resources [on reddit](https://www.reddit.com/r/rust/comments/cpt4b7/how_to_start_with_rendy/).

FuriouZz:
What about [wgpu-rs](https://github.com/gfx-rs/wgpu-rs) ? Too high level?

mitchmindtree:
We haven't taken a close look at `wgpu` just yet, though we are aware of it! A couple of the reasons I'm interested in investigating `rendy` first:

1. Rendy's [rendergraph](https://github.com/amethyst/rendy#rendergraph) is a high-level abstraction we're very interested in, to the point where we have discussed implementing something similar ourselves in the past #243.
2. Web isn't the *main* priority for rendy, but it is still supported. We want to avoid getting too invested in a web-first platform as many of nannou's applications involve native installations with multiple windows across multiple displays (e.g. projection mapping with quad warping, triple screen setups, etc). It maybe very well be the case that this is possible with wgpu, but we haven't looked closely enough to be sure yet, and at least rendy seems generalised enough that it would be feasible for us to address support for this there, whereas it may be more difficult to motivate that kind of support in a render engine designed for the web first.

That said, there's a good chance both of these are also addressed by wgpu! If that's the case I'd love to hear about it - I just haven't personally had the time to investigate just yet.

mitchmindtree:
Thanks for your interest @kvark!

> For a project focused on "creative coding", you seem to be selecting a surprisingly challenging path for the graphics stack (Vulkano, Rendy's render graph, etc)

Hopefully I clarified how we ended up with vulkano in #408!

[...]

PR #396, opened 2019-09-15
Add the method `dispatch` to AddComands for AutoCommandBufferBuilder

mft-ableton:
I am interested in rendering the output of some compute shaders to the swap chain.
I added the `dispatch()` method to `AddCommands` related to `AutoCommandBufferBuilder`, so that compute pipeline execution and graphics rendering can take place in the same command buffer.
I also added an example to show one use case.

I saw the comment `Note that certain builder methods are *not* provided in order to avoid unexpected behaviour`  in `raw.rs`... was this one intentionally left out?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -147,6 +147,9 @@ path = "examples/vulkan/vk_images.rs"
 name = "vk_image_sequence"
 path = "examples/vulkan/vk_image_sequence.rs"
 [[example]]
+name = "vk_compute_shader_rendering"
+path = "examples/vulkan/vk_compute_shader_rendering.rs"
+[[example]]
 name = "vk_compute_shader"
 path = "examples/vulkan/vk_compute_shader.rs"
 [[example]]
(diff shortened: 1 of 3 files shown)


mitchmindtree:
Sorry about the delay @mft-ableton!

I think this is totally fine to add and makes a lot of sense. Thanks a lot for the new example as well!

Just a heads up, there's a good chance the `vk` module will change a lot (or eventually be removed) in the future in favour of a switch to `rendy` - see issue #374 where we've been discussing this a little. That said, I think this is super useful in the meantime, and we don't really have any estimate on time frame for that rendy work just yet anyway. Thanks!

mft-ableton:
Great! Thank you very much, I'm happy to contribute something that is not about a typo for once. Hehe.
I just read the the rendy issue. I have to say I am very happy with Vulkano so far but I'm still very fresh, so I'm more learning about what it can rather than what it can't.
What do you mean with slack? I was not aware I could communicate on slack with people outside of my company.

JoshuaBatty:
Hi @mft-ableton , see [this link](https://communityinviter.com/apps/nannou/nannou-slack) to get an invitation to our nannou slack channel. A lot of us hang out there all day and it's a fun place to ask questions and share nannou knowledge.

mft-ableton:
Ah great, I was not aware of it. I just registered.

PR #398, opened 2019-09-16
New Generative Design and NOC examples

JoshuaBatty:
This PR adds in the following new examples 

Generative Design - Random & Noise
* m_1_1_01
* m_1_2_01
* m_1_3_01

NOC - Oscillation Figures
* 3_08_static_wave_lines
* 3_09_exercise_additive_wave
* 3_10_pendulum_example_simplified
* 3_10_pendulum_example
* 3_11_spring
* 3_11_exercise_additive_wave
* 3_16_exercise_springs
* 3_16_exercise_springs_array
* 3_exercise_spring_sine
* 3_extra_oscillating_body

NOC - Systems
* 4_01_single_particle_trail
* 4_03_exercise_moving_particle_system
* 4_05_particle_system_inheritance_polymorphism
* 4_06_exercise_shatter
* 4_06_particle_system_forces
* 4_07_particle_system_forces_repeller
* 4_10_exercise_particle_intersection
* 4_10_exercise_particle_repel

NOC - Agents
* 6_01_seek_trail
* 6_02_arrive
* 6_03_stay_within_walls
* 6_03_stay_within_walls_trail

NOC - CA
* 7_game_of_life_wrap_around
* 7_hexagon_cells

NOC - Neural Networks
* 10_01_simple_perceptron
* 10_02_seeking_neural
* 10_03_network_viz

@mitchmindtree example 10_04_network_animation is the only one that isn't finished yet, not sure what the best idiomatic rust way would be to represent this code....?


examples/nature_of_code/chp_10_neural_networks/10_04_network_animation.rs
```diff
@@ -0,0 +1,253 @@
...
+    network.connect(&mut d, &e, random_f32());
+    network.connect(&mut e, &f, 1.0);
+
+    // Add them to the network
+    network.add_neuron(a);
+    network.add_neuron(b);
+    network.add_neuron(c);
+    network.add_neuron(d);
+    network.add_neuron(e);
+    network.add_neuron(f);
```
mitchmindtree: One possible solution would be to add the neurons *before* connecting them, and have the `add_neuron` method return the index of the neuron. Then have the `connect`method accept neuron indices (rather than references to neurons).  Does that fix the ownership issue you were running into?

JoshuaBatty: Yeah cool that makes sense but that not where my problem is. Its the fact that the "Network" has a vector of Neurons and a vector of Connections. Yet the connection struct also has needs references to two Neurons, and each Neuron needs to hold a vector of references to Connections! Phew. messy ugly stuff.....

mitchmindtree: Ahh yep I see what you mean!

The Processing approach is disallowed in Rust by the borrow checker for a pretty good reason. E.g. it would be very easy to remove some of the connections while forgetting to also remove them from the nodes involved in those connections (or vice versa).

The reason why Processing can allow this and be memory safe is because each of the Neurons and Connections are heap-allocated objects that will get garbage collected once they are no longer referenced. This means that a neuron accessing a removed connection would still be technically memory safe. That is, the object would still be there on the heap as the neuron still owns a reference to it (even if the user has removed it). That said, this would still likely result in broken, confusing logic due to the neuron trying to update with connections that should no longer exist. At the very least there would be a memory leak.

If we wanted to match this Processing approach more closely, we could create each node and each connection as `Rc<RefCell<Node>>` and `Rc<RefCell<Connection>>` respectively, but that would probably be even more confusing to new users!

I think the more rustic approach, or a safer approach in any language, would be to make it the responsibility of the `Network` to apply the update that occurs between neurons and their connections as the `Network` is the one true owner of all the nodes and all the connections, rather than making it the responsibility of each of the neurons and connections as the web of ownership gets messy when they all refer to each other.

I wonder if we can refactor out some of the self-mutating methods on neuron and connection and move that into the network's update method? If the Connection type just stores indices into the Neurons list (rather than references to neurons) and we remove the list of connections from the Neuron type (as we already know what their connections are via the connections list), we should be able to achieve this. I haven't looked too closely just yet, but this is probably how I'd approach this!

JoshuaBatty: Great, this sounds like a good approach. I was wondering whether we should start using RefCell's but agree, it probably makes the nature of code examples a bit hairier than they need to be. I'll have a go at refactoring the code using indices in place of references to types and have the Network type handle all the logic. Cheers for the feedback.

JoshuaBatty:
Going to merge this now so they are ready for the RustFest workshop tomorrow, still need to work on 10_04_network_animation so i'll leave that example out for the time being.

PR #411, opened 2019-09-26
Made simple_loop example

MacTuitui:
A simple example that shows how to do a loop using noise for example.

### DIFF
--- a/examples/simple_loop.rs
+++ b/examples/simple_loop.rs
@@ -1,0 +1,100 @@
+/*
+ * Simple Loop example with 4D noise
+ * Daily Sketch 2019/09/23 by Alexis Andre (@mactuitui)
+ *
+ */
+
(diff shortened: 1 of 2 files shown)


examples/simple_loop.rs
```diff
@@ -0,0 +1,100 @@
...
+        .build()
+        .unwrap();
+    let mut noise = Perlin::new();
+    noise = noise.set_seed(1);
+    Model { noise }
+}
+
+fn update(_app: &App, _model: &mut Model, _update: Update) {}
+
+fn window_event(_app: &App, _model: &mut Model, event: WindowEvent) {
```
JoshuaBatty: Hi @MacTuitui thanks heaps for this example. Can I suggest that we remove the window event and update methods from this example? Think it would be good to keep these *simple* examples as concise as possible.

Issue #180, opened 2018-07-20
Allow for accepting Rgb Color values via their byte representation.


freesig:
When creating a Rgb value a user might assume they are 0-255 not 0.0 to 1.0. 
Currently the sketch will just compile. This could lead to confusion.

There's a few options. 
1. We could automatically scale the value from 0-255 to 0.0-1.0
2. Emit a warning and help suggestion.
3. assert and stop compile.
4. panic! at runtime.

Personally I like a combo of 1. and 2. because it allows the user to continue on but they are still aware.

mitchmindtree:
Yeah this is a good point - the current behaviour might be surprising for folks used to the way that processing handles colours! That said, I think there'd be an equal number of folks just as confused if we were to go in the other direction.

openFrameworks [seem to handle this](https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/types/ofColor.h#L626-L636) by having different implementations depending on whether the `ofColor` is constructed with an `integer` or a `float`. It looks like [Cinder](https://github.com/cinder/Cinder/blob/master/include/cinder/Color.h) also have a similar solution.

We should also consider that the user has a little less precision if we decide to expose a byte-per-channel API like processing. Integer values can also make interpolating colour channels a bit more difficult, as integers aren't accurately interpolate-able via a single scalar value with stepping.

Perhaps we could provide an alternate constructor? Like `.rgb8(0, 123, 213)` or `Rgb8`.

Would be sweet if we could get some more opinions on how to go forward! It's tricky trying to strike this balance between simplicity and correctness :)

Oh also, currently we're using the [palette](https://docs.rs/palette/0.2.0/palette/) crate for colours behind the scenes, however we're currently 2 versions behind and I think there's been some pretty major breaking changes. I wonder if they support byte-per-channel representations.

freesig:
Yeh an alternate constructor combined with an assert check if someone enters > 1.0 into the original constructor. 
I assumed it was 0-255 and then plugged 255.0 into `Rgb::new(255.0, 255.0 255.0)`
So we need to tell the user that the range is 0.0 - 1.0
I think if we have an alternate constructor for people that prefer thinking is 0-255 values then it should definitely fail to compile if they accidentally use the 0.0 - 1.0 constructor.

mitchmindtree:
If anyone is interested in implementing this, I think this can be nicely addressed by:

- [ ] Decide whether the function should be called `rgb8` or `rgb_bytes` or something else.
[...]

PR #413, opened 2019-10-02
added feature to use .rgb8(u8,u8,u8) as an option to color stuff

linusbeff:
also `rgb8(u8, u8, u8)` has been added as a shortcut to `Rgb::<u8>::new()`

#180


mitchmindtree:
Thanks so much for the PR @linusbeff, and so sorry for the delay in response! We've since added an `rgb8` constructor but don't have the types as you do here. I'll rebase your PR now and open a new one :+1:

PR #414, opened 2019-10-07
Remove `secondary` from methods of the command buffer builder

mft-ableton:
Remove the argument `secondary` from the methods `begin_render_pass()` and `next_subpass()` of the command buffer builder `AddCommands`.
This change is motivated by the fact that `AddCommands` doesn't offer the possibility to use secondary command buffers, which means that `secondary` had to be always set to `false`.

### DIFF
--- a/examples/vulkan/vk_compute_shader_rendering.rs
+++ b/examples/vulkan/vk_compute_shader_rendering.rs
@@ -169,7 +169,7 @@ fn view(app: &App, model: &Model, frame: &Frame) {
             push_constants,
         )
         .expect("failed to add `dispatch` command")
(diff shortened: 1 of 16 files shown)


mitchmindtree:
Thanks so much @mft-ableton, this looks good! I think travis is complaining about some minor formatting - would you mind running `cargo fmt`? We still need to add a github action or something to the repo that automatically adds a formatting commit for us :)

mft-ableton:
Sure, I forgot about it. It's done now :)

mitchmindtree:
Looks good to me!

Looks like nightly is still failing due to rustfmt being unavailable on the nightly toolchain atm, but that's unrelated and looks like it may get addressed in #415. Going to merge.

PR #415, opened 2019-10-08
Use a separate job for validating docs

yutannihilation:
(Disclaimer: I'm a very newbie to Rust. I'm sorry if my understanding is wrong...)

Since it seems `cargo doc` is a separate thing from `cargo test`, I think it can be moved to a separate job to reduce the time per job so that it'll be less likely to be timed out.

Besides, I added `--no-deps` option. If I understand correctly, it's not necessary to generate documents for all the dependencies just to test the nannou's doc is correct, right?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,10 @@ matrix:
               - cargo doc --no-deps -v
 
 script:
-    - cargo fmt --all -- --check
+    - |
+      if [[ "$TRAVIS_RUST_VERSION" == stable ]]; then
+        cargo fmt --all -- --check
+      fi
     - cargo build -v
     - cargo test -v
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,4 @@
 language: rust
-before_script:
-- rustup component add rustfmt
 
 addons:
     apt:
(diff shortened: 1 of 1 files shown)


mitchmindtree:
Thank you so much @yutannihilation, this is a really big help!

While we're tweaking travis, would it also be possible to move the `cargo fmt --all -- --check` line under the `stable` job, so that it only runs for the stable branch? Sometimes the `rustfmt` component is unavailable for the nightly branch, and I think this would avoid us running into errors (as it seems is currently the case). Alternatively, maybe we can move it under its own job altogether so that it is easier to identify when travis fails due to incorrect formatting? Either way I don't mind, I'll let you decide!

yutannihilation:
Thanks! I added a if condition. How about this?

stable: https://travis-ci.org/nannou-org/nannou/jobs/595106346#L231-L233
beta: https://travis-ci.org/nannou-org/nannou/jobs/595106345#L231-L233

mitchmindtree:
Ahh yes this looks good!

Is it possible to also make sure the `rustup component add rustfmt` line under the `before_script:` section also only happens on the `stable` job? It looks like travis is failing to add the `rustfmt` component on the `nightly` job, but we only really care about the `stable` job.

yutannihilation:
Ah, I see! So, I think the possible choices are:

1. Move `before_script` under `stable` job of matrix
2. Use the same if condition on `before_script`

and I chose option 2 because it seems clearer to put `before_script` near `script`.

mitchmindtree:
Perfect! Thanks again :)

yutannihilation:
Thanks!

PR #424, opened 2019-11-11
Add simple_ui_image Example

### LINKED ISSUE #316: image example
Hi there,

the README claims that images can be loaded (via the GUI). So far I failed to see how. An example (may as extension to the `simple-ui` example) would be useful for beginners. See also #155 

Cheers,
tpltnt

tuomas56:
This adds an example for how to display images in the ui framework using the `primitive::image::Image` widget. This would resolve #316. I also have a change that would put most of the meat of this example into a method on the `Ui` struct itself (tentatively called `upload_image`), making this a lot simpler:
```rust
// Generate the image path, and load the image using the image crate.
let logo_path = app.assets_path().unwrap().join("images").join("Nannou.png");
let image = nannou::image::open(logo_path).unwrap();

// Upload it to the Ui.
let image_id = ui.upload_image(app, &image).unwrap();
```
But, given that current image API is not permanent (?), I'm not sure if I should submit a pull request for that.

### DIFF
--- a/examples/simple_ui_image.rs
+++ b/examples/simple_ui_image.rs
@@ -1,0 +1,61 @@
+use nannou::prelude::*;
+use nannou::ui::prelude::*;
+use nannou::conrod_vulkano;
+
+fn main() {
+    nannou::app(model).update(update).simple_window(view).run();
+}
+
+struct Model {
+    ui: Ui,
+    widget_id: widget::Id,
+    image_id: image::Id
+}
+
+fn model(app: &App) -> Model {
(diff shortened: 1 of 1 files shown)


mitchmindtree:
Hi @tuomas56, so sorry for the delay in response to this, this one totally slipped me by!

We recently made the switch from vulkano to wgpu for our graphics and conrod backends, so I think this will need a rebase. I'll try to remember to take a look into doing so soon in the case no one beats me to it!

PR #432, opened 2019-11-27
Adding rgb8 to represent values with bytes

### LINKED ISSUE #180: Allow for accepting Rgb Color values via their byte representation.

DavidPartouche:
closes  #180

There's gonna be a merge conflict with the other PR I've pushed (PR #431)
I'll fix it depending on which one you merge first.


mitchmindtree:
Thanks again for these @DavidPartouche! I've merged the gray PR first so this one should be ready to be resolved

PR #434, opened 2019-12-02
Fixed compiler error

adamnemecek:
I was getting a compilation error. I updated cgmath to version 0.17 (from 0.16) and approx to 0.3.2 (from 0.1). 

There are some changes to use `Self` instead of `$VectorN<S>` where possible.


tpltnt:
Can is PR be closed in light of #437

mitchmindtree:
I believe so, thanks!

PR #437, opened 2019-12-22
Update dependencies

oherrala:
Two commits:

 *  Update cgmath 0.17 and remove approx crate
     cgmath crate publishes three traits from approx 0.3 crate (`AbsDiff`, `RelativeEq`, `UlpsEq`) which are also implemented in this commit while removing the old ApproxEq implementation from approx 0.1 crate.
 * Update everything else. No code changes required here.


oherrala:
Apparently `cgmath` and `approx` parts are also changed in #434, but that PR also seems to change many other things.

mitchmindtree:
Thanks for this @oherrala!

I think there's one line of formatting that travis is unhappy about - would you mind running `cargo fmt`?

oherrala:
@mitchmindtree Sorry about that, cargo fmt done.

tpltnt:
looks good to me.

PR #441, opened 2020-01-04
Fix simple audio example

tpltnt:
Hi there,

I adjusted the simple audio example to not underrun (only pressing down results in a frequency increase at some point) or overrun (since f64 is way larger than the human hearing range).

Cheers,
tpltnt

### DIFF
--- a/examples/simple_audio.rs
+++ b/examples/simple_audio.rs
@@ -68,7 +68,13 @@ fn key_pressed(_app: &App, model: &mut Model, key: Key) {
             model
                 .stream
                 .send(|audio| {
-                    audio.hz += 10.0;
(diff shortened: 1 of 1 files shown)


mitchmindtree:
Thanks for the PR @tpltnt! I think this unnecessarily complicates the example a little though. The idea is to show in a simple-as-possible manner how the user can update some state on the audio thread from the GUI thread. I'd personally prefer that we leave it as is for now. Have this overflow/underflow caused you an issue in practise? Or is this more of a pre-emptive fix?

tpltnt:
I have not observed an overflow problem yet. The underflow actually occurs when you only press down ... or at least the audio changes in an unexpected way.

Issue #448, opened 2020-02-15
Update `nannou-new` to work with cargo workspaces


mitchmindtree:
This could be done by first checking for a `Cargo.toml` containing a `[workspace]`. If one is found, it could be updated to include the new crate before the directory is created.

kneitinger:
Interestingly enough, workspaces have broken `nannou-new`.

In d6d5d38a6b617cf976f857b79c3c8bb233a22aad, the examples directory was moved from `nannou/nannou/examples` (the nannou package in the nannou repo) to `nannou/examples` - its own package in the repo.

In addition to `nannou-new` not being able to find `template_app.rs` and `template_sketch.rs` because of their new location, there is another issue, which I cannot figure out how to fix just in code.

---
_(note: the hypothetical steps in following description do not match `nannou-new`'s current code, since it's current approach of grabbing the most recent `nannou` crate to both figure out version and grab the templates, can no longer be used)_

The problem stems with the lack of versioning on the repo as a whole. Say we query crates.io to find the most up to date `nannou` release so that we can make an entry in our new project's `Cargo.toml`.  That works fine, and we now know we're targeting `0.14.1`.  Now how do we find the relevant templates?  We cannot use `cargo` since the `examples` crate is rightfully unpublished.

At this point we have to grab it from github directly, which I think is acceptable, however, there is currently not a way to determine the proper version of the template file to grab.

If we naively clone the `examples/templates/template_app.rs` file from `master`, we run the risk of generating projects that do not compile, if for example, the template app file is modified to include an exciting new feature that hasn't hit crates.io yet.

Currently, the `nannou` repo does not use any tagging or releases.  This means that it is very difficult to correlate a version published on crates.io to the commit at which it was created.

---
What I see as the best solution (but I'm sure there are plenty that I haven't thought of) is for this repo to start tagging commits that we're pushed to crates.

Because this repo is organized as a workspace, the semantics of tags are not perfectly obvious, but two solutions could work out well:

1. Create a tag for every version of the `nannou` crate, since it arguably impacts most every crate.  So for instance, if `nannou` 0.14.2 came from commit `39c9ed`, then the git tag `0.14.2` or `v0.14.2` would point to that commit.  This is pretty low maintenance, but might be confusing to have a workspace versioned in such a way.
[...]

PR #457, opened 2020-02-25
[nannou-new] Fix build with 2018 edition

dtamai:
I was looking into #448 so I tried to build the `nannou-new` binary but it is failing because of the 2018 edition changes, it can't find the `cargo` crate.

### Changes

- set the edition to 2018 in the _Cargo.toml_ file.
  This is the required change to fix the build
- update dependencies and make the required changes
- fix issue with downloading a random version of the `nannou` package
  I didn't investigate if this happens with the previous cargo version, but in the 0.42 the `query` method doesn't yield a package summary in any order, so the last one that is downloaded could be any version.

### DIFF
--- a/nannou-new/Cargo.toml
+++ b/nannou-new/Cargo.toml
@@ -8,8 +8,9 @@ keywords = ["tool", "build", "nannou", "creative", "sketch"]
 license = "MIT"
 repository = "https://github.com/nannou-org/nannou.git"
 homepage = "https://github.com/nannou-org/nannou/tree/master/nannou-new"
+edition = "2018"
 
 [dependencies]
-cargo = "0.28" # For retrieving the latest `nannou` version and working with cargo-clone.
+cargo = "0.42" # For retrieving the latest `nannou` version and working with cargo-clone.
 names = "0.11" # For generating random project names.
-rand = "0.4" # For generating random beverages (very important).
+rand = "0.7" # For generating random beverages (very important).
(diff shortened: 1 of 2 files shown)


mitchmindtree:
Thanks so much for this @dtamai! This is a great help :)

Also, something that seems much clearer in retrospect, is that we should probably call into the user's existing installation of `cargo`, rather than use the `cargo` crate itself - this would greatly improve `nannou-new` compile times and also avoid an issue that sometimes occurs where the package needs to be built twice as a result of the separate cargos (the first build by the cargo dep, the second in the user's environment). Anyways, this is best left for a separate PR!

dtamai:
> Also, something that seems much clearer in retrospect, is that we should probably call into the user's existing installation of `cargo`, rather than use the `cargo` crate itself - this would greatly improve `nannou-new` compile times and also avoid an issue that sometimes occurs where the package needs to be built twice as a result of the separate cargos (the first build by the cargo dep, the second in the user's environment). Anyways, this is best left for a separate PR!

It would be nice! Without the `cargo` crate it should compile much faster.

I could find the `nannou` version and template files using just the `cargo` tool from the command line, so it should be doable from the script, using `jq` for JSON parsing.

First it needs a new package with `nannou` as dependency:

```bash
cargo search nannou --limit 1
```

This prints a string ready to be used in the dependencies table so the first requiremente is done. Then:

```bash
cargo fetch # to download the package
cargo metadata --format-version=1 \
    | jq -r '.packages[] | select(.name == "nannou") | .manifest_path'
```

This gives the location of the `Cargo.toml` file for the `nannou` package installed in the local Cargo cache. I couldn't find another way to make cargo download the package and tell where the source code is.

I think I can change the script to use the cargo command instead of the crate.

PR #488, opened 2020-03-24
Window doc fix: explicit tuple

tpltnt:
Hi there,

I made the meaning of "size" (for windows) more explicit by refering to width and height and thus explaining the return values. (Size itself might also imply the area.) No code changes.

Cheers,
tpltnt


mitchmindtree:
Thanks!

PR #493, opened 2020-03-26
Adding in a ton of new generative design examples before the restructure 

JoshuaBatty:


mitchmindtree:
No travis, but tested locally und alles gut!

PR #495, opened 2020-03-28
Fix "redraw event in non-redraw phase"

schulzch:
This PR fixes "redraw event in non-redraw phase" on window resize/move by updating to winit 0.22. See [winit #1382](https://github.com/rust-windowing/winit/issues/1382).


mitchmindtree:
Thanks for this! Travis seems to be out of action atm, but I've tested locally and all seems fine :+1:

PR #503, opened 2020-04-01
Fix comments for the generation of SPIR-V files in wgpu examples

mft-ableton:
Fix names of the generated SPIR-V files in the comments (which were all invariably `vert.spv`) and remove the `-o ...` argument since it corresponds to the default pattern for the name of the generated file.


mitchmindtree:
Thanks heaps! I knew I would mess this up somewhere with the amount of copying and pasting of this comment, but not how much :woman_facepalming:

PR #507, opened 2020-04-02
Add function wgpu::shader_from_spirv_bytes()

mft-ableton:
wgpu::shader_from_spirv_bytes() reduces the amount of code required for loading shaders.

### DIFF
--- a/examples/wgpu/wgpu_compute_shader/wgpu_compute_shader.rs
+++ b/examples/wgpu/wgpu_compute_shader/wgpu_compute_shader.rs
@@ -41,9 +41,7 @@ fn model(app: &App) -> Model {
     let device = window.swap_chain_device();
 
     // Create the compute shader module.
-    let cs = include_bytes!("shaders/comp.spv");
-    let cs_spirv = wgpu::read_spirv(std::io::Cursor::new(&cs[..])).unwrap();
-    let cs_mod = device.create_shader_module(&cs_spirv);
+    let cs_mod = wgpu::shader_from_spirv_bytes(device, include_bytes!("shaders/comp.spv"));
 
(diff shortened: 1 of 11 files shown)


mitchmindtree:
Thanks heaps for the PR @mft-ableton!

Nice work on your first macro ever! They can be super handy, like super powers :)

In general, I try to keep macro usage to a minimum for a couple reasons:

- They can be a little scary for newer rust users interested in contributing.
- They can result in quite a bit of extra code generation in cases where a function could suffice.

That said, I think you're totally on the right track attempting to reduce this code duplication! What are your thoughts on something like this, where `shader_from_spirv_bytes` is a helper function we expose from the `wgpu` module?

```rust
let cs_mod = wgpu::shader_from_spirv_bytes(device, &include_bytes!("shaders/comp.spv"));
```

This way we can still get it down to a one liner, but perhaps with a little less macro magic? It also has the nice benefit of supporting loading spirv bytes loaded at runtime too.

mitchmindtree:
Oh also

> let me know if the size of the commit is ok

This size is totally fine!

mft-ableton:
> What are your thoughts on something like this, where `shader_from_spirv_bytes` is a helper function we expose from the `wgpu` module?
> 
> ```rust
> let cs_mod = wgpu::shader_from_spirv_bytes(device, &include_bytes!("shaders/comp.spv"));
> ```

This makes a lot of sense to me, minimal change with maximal effect :)
I'll do that.

mft-ableton:
@mitchmindtree done! Also I realised that `&` or `&[]` is not required with the result of `include_bytes!()`, I guess because it is a reference to a static array.

mitchmindtree:
Perfect! Looks great, thanks again :)

PR #511, opened 2020-04-03
Add random_ascii() function to rand module

JoshuaBatty:
now allows to easily generate a random ascii character. The possible characters that can be returned are: 

"ABCDEFGHIJKLMNOPQRSTUVWXYZ
 abcdefghijklmnopqrstuvwxyz
 0123456789)(*&^%$#@!~. "

The nature of code 9_1_ga_shakespeare example has also been updated to use this function now. @mitchmindtree should I be declaring this asciiset as a const inside the function or is it better if this lives at the top of the rand module?


mitchmindtree:
Looks good to me!

PR #514, opened 2020-04-05
Move changelog into the guide. Remove old travis scripts.

mitchmindtree:


mitchmindtree:
@JoshuaBatty I think the last thing for the guide migration now is to add those fancy netlify CI things that allow you to see a preview of the rendered doc :) do you ;remember how to set this up?

PR #530, opened 2020-04-10
changed `color_points` to `points`

paulgb:
It seems the example in `drawing-2d-shapes.md` is broken in the latest version. This fixes it.

I am curious, though, if there is an equivalent to the behaviour of `colored_points` to give each line segment a different colour?


mitchmindtree:
Hi thanks for this @paulgb!

Ahh it looks like these code snippets aren't enabled for automated testing so they must have slipped through the cracks!

For a sequence of colored points, the convention has changed from `color_points` to `points_colored` - this is the case for each of `draw.mesh()`, `draw.polyline()` and `draw.polygon()`. Would you mind updating this PR to use `points_colored` instead?

paulgb:
Thanks for the explanation. I will test it and update it now.

PR #533, opened 2020-04-11
Fix incorrect behavior of .z_radians()

### LINKED ISSUE #532: Rotating a polygon with .z_radians() causes an incorrect rotation
Under nannou 0.13.1, calling .z_radians() to rotate a polygon causes it to be rotated in the y-axis instead of the z-axis.

Here is an application to replicate this behavior:
```rust
use nannou::prelude::*;

fn main() {
    nannou::sketch(view).size(400, 150).run();
}

fn view(app: &App, frame: Frame) {
    let draw = app.draw();

    draw.background().color(BLACK);

    let points = [(-30., 0.), (-10., 30.), (30., -30.)];
    let angle = app.time;

    draw.polygon()
        .x(-150.0)
        .x_radians(angle)
        .points(points.iter().copied());
    
    draw.polygon()
        .x(0.0)
        .y_radians(angle)
        .points(points.iter().copied());

    draw.polygon()
        .x(150.0)
        .z_radians(angle) // This causes the same effect as y_radians()
        .points(points.iter().copied());

}
```

dzil123:
Fixes #532

### DIFF
--- a/nannou/src/draw/drawing.rs
+++ b/nannou/src/draw/drawing.rs
@@ -439,7 +439,7 @@ where
 
     /// Specify the orientation around the *z* axis as an absolute value in radians.
     pub fn z_radians(self, z: S) -> Self {
-        self.map_ty(|ty| SetOrientation::y_radians(ty, z))
+        self.map_ty(|ty| SetOrientation::z_radians(ty, z))
     }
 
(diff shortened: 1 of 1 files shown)


mitchmindtree:
Thanks so much!

PR #534, opened 2020-04-11
add note about message being popped

madskjeldgaard:
Sorry, I did this before my morning coffee so no seperate branch but hope it's ok for a small documentation change


mitchmindtree:
Of course! thanks again :)

Issue #323, opened 2019-05-26
Create Color from (hex) Integers


rylev:
It would be nice to be able to convert hex integers into colors so the one could do the following:

```rust
draw.ellipse()
            .x_y(x, y)
            .radius(0.5)
            .color(0x2ECC40)
```

MacTuitui:
I am also for it, but in the mean time, this is how I do the manual conversion if it helps (disclaimer: I'm still a beginner so there might be better ways):

```rust
let colors: [u32; 13] = [0xFF70382E,0xFF48221F,0xFF25242C,0xFF5B413F,0xFF2C474C,0xFF306C6F,0xFF555656,0xFF288C8E,0xFF0AF7F7,0xFF1FAEAF,0xFF19D3D3,0xFF587473,0xFFB2DBDD];
let colorsv = colors.to_vec();
let colsrgb : Vec<Rgb> = colorsv.into_iter().map(|c| {  
    let blue: u8 = (c & 0xFF) as u8;    
    let green: u8 = (( c >> 8 ) & 0xFF) as u8;
    let red: u8 = (( c >> 16 ) & 0xFF) as u8;
    let c = Srgb::new_u8(red, green,blue);
    c.into()
}).collect();
```

Then I have an array of colors that nannou accepts (or that you can `into()` if needed).

cronokirby:
Doesn't the following function kind of satisfy this?
https://docs.rs/nannou/0.11.1/nannou/ui/prelude/color/fn.rgb_bytes.html

then you could do `.color(rgb_bytes(0x2E, 0xCC, 0x40))`
you could also add similar functions that take a single u32, e.g. `rgb_u32`

jli:
I think this has been fixed by #545?

mitchmindtree:
Closed via #545

PR #545, opened 2020-04-16
add rgb_u32 to create color from hex literal

paulkernfeld:
implements #323

### DIFF
--- a/nannou/src/color/mod.rs
+++ b/nannou/src/color/mod.rs
@@ -6,7 +6,6 @@ pub mod conv;
 
 pub use self::conv::IntoLinSrgba;
 pub use self::named::*;
-use crate::ui::color;
 #[doc(inline)]
 pub use palette::*;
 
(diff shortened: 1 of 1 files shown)


paulkernfeld:
Hey! This is my first PR so I'm not that familiar with the organization of nannou's code. Please let me know if there are any changes I should make so that this fits in more smoothly 😁

nannou/src/color/mod.rs
```diff
@@ -105,6 +106,15 @@ where
     LinSrgba::new(r, g, b, a)
 }
 
+/// Create a new color from a hexadecimal int literal
+#[inline]
+pub fn rgb_u32(c: u32) -> color::Color {
+    let blue: u8 = (c & 0xFF) as u8;
+    let green: u8 = ((c >> 8) & 0xFF) as u8;
+    let red: u8 = ((c >> 16) & 0xFF) as u8;
+    color::rgb_bytes(red, green, blue)
```
mitchmindtree, line 115: Would you mind using the `rgb8` function that is in this module, rather than the `ui::color::rgb_bytes` function here? E.g.
```rust
rgb8(red, green, blue)
```
This should also allow for removing the `use crate::ui::color` import. You may also need to change the return type to `Rgb<u8>`.

paulkernfeld: No problem, just pushed a change that should fix this 😀

mitchmindtree:
Perfect! Thanks again :)

PR #548, opened 2020-04-18
Add osc sender guide + moving circle guide

madskjeldgaard:
This adds guides for making a circle which moves about on the screen (example of map_range and using sine waves). 

It also adds guides for setting up OSC Sender. 

I will make one for a OSC receiver guide as well when I get the time.


madskjeldgaard:
The gif in the moving-a-circle guide looks terrible. If anyone can do a better job at producing a better gif then go ahead :)

madskjeldgaard:
Oh, and I see that it won't build. Is this due to a lack of `norun` in the code bits?

guide/src/tutorials/basics/moving-a-circle-about.md
```diff
@@ -0,0 +1,119 @@
...
+```
+Let's make another one but at half the speed by dividing the time value by two
+```rust
+let slowersine = (app.time / 2.0).sin();
+```
+Now we have two functions generating nice, smooth wave movements and we want to control our little circle using these.
+
+But we if we put these values in the (x,y)-coordinates of the `.x_y()`-method we used before we would not see much movement. That's because the sine waves generate values between -1.0 and 1.0 and the coordinates expect a pixel position.
+
+But how wide is the window the circle has been drawn in ? To get a precise idea of this, we can use a handy method called [window_rect](https://docs.rs/nannou/latest/nannou/app/struct.App.html#method.window_rect) which is available in the `app` variable.
```
mitchmindtree, line 51: Maybe rather than:

> But how wide is the window the circle has been drawn in ?

this could just be:

> But how wide is our window?

guide/src/tutorials/osc/osc-sender.md
```diff
@@ -0,0 +1,155 @@
...
+
+    // Clear the background to purple.
+    draw.background().color(PLUM);
+
+    // Draw a blue ellipse at the x/y coordinates 0.0, 0.0
+    draw.ellipse().color(STEELBLUE).x_y(x, y);
+
+    draw.to_frame(app, &frame).unwrap();
+}
+```
```
mitchmindtree, line 155: Maybe this could end with something like:

> In the [next tutorial](<link-to-osc-sender-tute>) we will take a look at how to receive our OSC values and do something interesting with them.

madskjeldgaard:
Okay, I think I've gotten through this. It was a lot of work to be honest dealing with strange errors, but a lot of them - it turned out - were due to a lack of whitespace between `#` and the commented out code (for example `fn main()`).

Things to take into consideration for others venturing into this:
* All examples used should be able to compile on their own. If you only want the reader to see parts of the code, comment out whatever you want to hide using the pound sign `#`
* Make sure there is whitespace between the pound sign `#` and the code you are commenting out
* Compiler warnings are considered errors here - for example unused variables or imports.

madskjeldgaard:
I think this is done @mitchmindtree .... pheeeeewww

mitchmindtree:
Perfect, great work @madskjeldgaard! Looking forward to the receiver tute :)

PR #552, opened 2020-04-20
Small guide fixes

jli:
This removes the Travis CI badge for the previous standalone guide repo. I didn't see a guide-specific workflow in the actions (https://github.com/nannou-org/nannou/actions). Lemme know if the guide readme should use the overall badge, or maybe there should be a guide-specific workflow?


mitchmindtree:
Thank you so much for this!

Ahh you are right, we don't have a dedicated workflow, but a few of the actions under the nannou workflow do test for the guide and publish new versions as it is updated. Would you recommend we create a separate workflow for guide-related actions?

Either way, happy to merge this, thanks again!

PR #561, opened 2020-04-22
Some documentation fixes

jli:
Added some clarifications to things I was wondering about (e.g. `map_range` not clamping output, `Resized` values in points, etc.).

Also fixes some broken internal links I found. (Maybe one day rust-lang/rust#43466 will land! In the meantime, I wonder if we could do a manual check that links in the generated output are valid? Might look into this, if you think it's worthwhile.)

### DIFF
--- a/nannou/src/app.rs
+++ b/nannou/src/app.rs
@@ -6,8 +6,6 @@
 //! - [**App**](./struct.App.html) - provides a context and API for windowing, devices, etc.
 //! - [**Proxy**](./struct.Proxy.html) - a handle to an **App** that may be used from a non-main
 //!   thread.
-//! - [**Draw**](./struct.Draw.html) - a simple API for drawing graphics, accessible via
-//!   `app.draw()`.
 //! - [**LoopMode**](./enum.LoopMode.html) - describes the behaviour of the application event loop.
 
 use crate::draw;
--- a/nannou/src/event.rs
+++ b/nannou/src/event.rs
@@ -1,9 +1,10 @@
 //! Application, event loop and window event definitions and implementations.
 //!
 //! - [**Event**](./enum.Event.html) - the default application event type.
-//! - [**winit::event::WindowEvent**](./struct.WindowEvent.html) - events related to a single window.
-//! - [**WindowEvent**](./struct.WindowEvent.html) - a stripped-back, simplified,
-//!   newcomer-friendly version of the **raw**, low-level winit event.
+//! - [**winit::event::WindowEvent**](https://docs.rs/winit/0.22.1/winit/event/enum.WindowEvent.html) -
+//!   events related to a single window.
+//! - [**WindowEvent**](./enum.WindowEvent.html) - a stripped-back, simplified, newcomer-friendly
+//!   version of the **raw**, low-level winit event.
 
 use crate::geom::{self, Point2, Vector2};
 use crate::window;
--- a/nannou/src/event.rs
+++ b/nannou/src/event.rs
@@ -1,6 +1,6 @@
 //! Application, event loop and window event definitions and implementations.
 //!
-//! - [**Event**](./enum.Event.html) - the defualt application event type.
+//! - [**Event**](./enum.Event.html) - the default application event type.
 //! - [**winit::event::WindowEvent**](./struct.WindowEvent.html) - events related to a single window.
 //! - [**WindowEvent**](./struct.WindowEvent.html) - a stripped-back, simplified,
 //!   newcomer-friendly version of the **raw**, low-level winit event.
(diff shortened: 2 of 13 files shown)


nannou/src/event.rs
```diff
@@ -1,9 +1,10 @@
 //! Application, event loop and window event definitions and implementations.
 //!
-//! - [**Event**](./enum.Event.html) - the defualt application event type.
-//! - [**winit::event::WindowEvent**](./struct.WindowEvent.html) - events related to a single window.
-//! - [**WindowEvent**](./struct.WindowEvent.html) - a stripped-back, simplified,
-//!   newcomer-friendly version of the **raw**, low-level winit event.
+//! - [**Event**](./enum.Event.html) - the default application event type.
+//! - [**winit::event::WindowEvent**](https://docs.rs/winit/0.22.1/winit/event/enum.WindowEvent.html) -
```
mitchmindtree, line 4: Would you mind changing:

> https://docs.rs/winit/0.22.1/winit/event/enum.WindowEvent.html

to

> https://docs.rs/winit/latest/winit/event/enum.WindowEvent.html

This will ensure we are always pointing to the latest version, provided winit doesn't move the type again :sweat_smile:

nannou/src/app.rs
```diff
@@ -6,8 +6,6 @@
 //! - [**App**](./struct.App.html) - provides a context and API for windowing, devices, etc.
 //! - [**Proxy**](./struct.Proxy.html) - a handle to an **App** that may be used from a non-main
 //!   thread.
-//! - [**Draw**](./struct.Draw.html) - a simple API for drawing graphics, accessible via
```
mitchmindtree, line 9: I wonder if we can keep this, but instead point to the `nannou::Draw` type (re-exported from the crate root)?

Some context in case you're interested - the `app` module used to have a thin wrapper `Draw` type that wrapped the "main" `draw::Draw` type. The idea was that it was a little simpler to use alongside the `app.draw()` method, but in the end we were able to simplify the main `draw::Draw` type enough that the `app::Draw` wrapper was no longer necessary. Looks like I forgot to update this path!

nannou/src/event.rs
```diff
@@ -40,8 +41,10 @@ pub struct Update {
 pub enum Event {
     /// A window-specific event has occurred for the window with the given Id.
     ///
-    /// This event is portrayed both in its "raw" form (the **winit::event::WindowEvent**) and its
-    /// simplified, new-user-friendly form **SimpleWindowEvent**.
+    /// The event is available as a **WindowEvent**, a more user-friendly form of
+    /// **winit::event::WindowEvent**. Once
+    /// [winit#1387](https://github.com/rust-windowing/winit/issues/1387) is fixed, its "raw" form
+    /// will also be available.
```
mitchmindtree, line 47: Amazing work for tracking this down, I hope this wasn't too confusing!

nannou/src/math.rs
```diff
@@ -18,7 +18,27 @@ pub use self::cgmath::{
...
+/// # use nannou::prelude::*;
+/// assert_eq!(map_range(3, 0, 10, 0.0, 1.0), 0.3);
+/// ```
+/// ```
+/// # use nannou::prelude::*;
+/// // When the value is outside the input range, the result will be outside the output range.
+/// let result = map_range(15, 0, 10, 0.0, 1.0);
+/// assert_eq!(result, 1.5);
+/// assert_eq!(clamp(result, 0.0, 1.0), 1.0);
+/// ```
```
mitchmindtree, line 41: :heart:

nannou/src/app.rs
jli: I'm happy to make that change (I was originally considering doing just that), but I'm wondering if it might be confusing? This is the documentation for the `app` module after all, and if there isn't a `app::Draw` type or an `app::draw` module, it seems potentially confusing to mention `Draw` at the top here?

Are you thinking that users might start off reading the `app` module docs as an entry point to learn about the API surface, and so it would be good to link to `Draw` from here? Or is it more about backwards compatibility for users who are used to the original module layout?

jli:
I noticed some more broken relative links... A general issue is that relative links included in the first line of a nested module's documentation comment will get included in the parent module, with a broken relative link.
 
Fixed by adding some line breaks :)

> Absolutely, this crosses my mind every time I write a link but I haven't taken the time to find a solution yet, if you were interested in tackling this that would be amazing!

As a quick hack, I wrote a Python script and it was finding many broken links, but I realized almost all of them were from other crates like conrod and palette. (Example: the "RGB" link in [nannou::color::Hsv](https://docs.rs/nannou/0.13.1/nannou/color/struct.Hsv.html) ("HSV is a cylindrical version of RGB") is broken, but that's due to palette's docs. See the same broken link on [palette:Hsv](https://docs.rs/palette/0.5.0/palette/struct.Hsv.html))

Ah, will need to think harder about this!

nannou/src/app.rs
mitchmindtree: Taking a closer look I totally agree! At a glance, for some reason I thought this section was a reference to `Draw` in relation to the `app.draw()` method, but now I see this is the top-level docs for the module :blush: Yes removing it like this is certainly the right change!

mitchmindtree:
Great stuff, thanks again @jli!

Perhaps if we come up with a nice crate-local solution, we can contribute some upstream PRs to those crates we expose publicly through nannou's API? I am the current maintainer for conrod, and the palette author has accepted contributions in the past :+1:

PR #572, opened 2020-04-25
Fix broken links on the index page of tutorials

yutannihilation:
It seems some of the links on this page won't work: https://guide.nannou.cc/tutorials.html

(Btw, the tutorials looks nicer than when I tried long ago! Thank you so much for your efforts.)


mitchmindtree:
Thanks so much @yutannihilation!

PR #576, opened 2020-04-27
fix implementation of angle_between

jli:
I don't think the current implementation of `(other - self).angle()` is correct. The angle between (1, 0) and (0, 1) should be PI, but this function gives 0.

I think `other.angle() - self.angle()` is what was meant.

Also fwiw, I took a look at the implementations in [Processing](https://github.com/processing/processing/blob/8e86389c7e017d0e4d61f81fb942c25e3ed348c7/core/src/processing/core/PVector.java#L987) and [p5.js](https://github.com/processing/p5.js/blob/1.0.0/src/math/p5.Vector.js#L1293). They both use the dot product / product of magnitudes formula, which i guess uses less trig operations and might be a faster? Maybe benchmarking this can be a separate issue though.

(Another note is that they both do clamping due to floating point issues, which I don't _think_ the difference-of-angles approach has? But probably also worthwhile testing this more thoroughly...)

### DIFF
--- a/nannou/src/geom/vector.rs
+++ b/nannou/src/geom/vector.rs
@@ -1402,19 +1402,21 @@ impl<S> Vector2<S> {
     /// ```
     /// # use nannou::prelude::*;
     /// # fn main() {
-    /// let a = vec2(-1.0, 1.0);
-    /// let b = vec2(1.0, 1.0);
-    /// assert_eq!(a.angle_between(b), 0.0);
-    /// assert_eq!(b.angle_between(a), PI);
-    /// assert_eq!(a.angle_between(b), (b - a).angle());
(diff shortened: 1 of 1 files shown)


jli:
Actually, this difference-of-angles approach doesn't work. Test case:

v1 = (cos(pi/4), sin(pi/4)  // v1.angle = pi/4 = 45°
v2 = (cos(pi + pi/6), sin(pi + pi/6)  // v1.angle = -5pi/6
v2.angle() - v1.angle() == -3.40

but angle_between should give a value between -pi and pi!

so I think we need to go with the dot product/product-of-magnitudes after all. will do a new commit.

mitchmindtree:
Thanks so much for digging into this @jli! Those tests make it so easy to review this too :) More than happy to merge!

PR #600, opened 2020-05-11
Example 1-5 Vector Magnitude: fix rect alignment

dominictwlee:
Rect should be aligned to top left of win. The old example drew half of the rect outside of win rect bounds.


mitchmindtree:
Perfect, thanks so much for the PR!

PR #601, opened 2020-05-11
change doc comments for lrgt fn to reflect tuple order

dominictwlee:
Just a minor change to the doc comments for the rect l_r_b_t function, so that it reflects the actual order of the tuple. I think this should resolve any ambiguities or confusion regarding the order of edges that get returned.


mitchmindtree:
Ahh nice spot, thanks so much for the PR!

Issue #430, opened 2019-11-25
Basic install fails - shaderc-sys is trying to use python 2.7


wefarrell:
On OSX 10.14.6 Ran:
```
cargo run --release --example simple_draw
```

Output:
```
   Compiling shaderc-sys v0.6.1
error: failed to run custom build command for `shaderc-sys v0.6.1`

Caused by:
  process didn't exit successfully: `/Users/willfarrell/dev/nannou/target/release/build/shaderc-sys-5a6f88a9d7cb3042/build-script-build` (exit code: 101)
--- stdout
cargo:warning=System installed library not found.  Falling back to build from source
running: "cmake" "/Users/willfarrell/.cargo/registry/src/github.com-1ecc6299db9ec823/shaderc-sys-0.6.1/build" "-DCMAKE_POSITION_INDEPENDENT_CODE=ON" "-DSPIRV_SKIP_EXECUTABLES=ON" "-DSPIRV_WERROR=OFF" "-DSHADERC_SKIP_TESTS=ON" "-DCMAKE_INSTALL_LIBDIR=lib" "-DCMAKE_INSTALL_PREFIX=/Users/willfarrell/dev/nannou/target/release/build/shaderc-sys-eb3c160072f82816/out" "-DCMAKE_C_FLAGS= -ffunction-sections -fdata-sections -fPIC -m64" "-DCMAKE_C_COMPILER=/usr/bin/cc" "-DCMAKE_CXX_FLAGS= -ffunction-sections -fdata-sections -fPIC -m64" "-DCMAKE_CXX_COMPILER=/usr/bin/c++" "-DCMAKE_BUILD_TYPE=Release"
-- Building SPIRV-Header examples
-- SPIRV-Tools: nosetests was not found - python support code will not be tested
-- Configuring incomplete, errors occurred!
See also "/Users/willfarrell/dev/nannou/target/release/build/shaderc-sys-eb3c160072f82816/out/build/CMakeFiles/CMakeOutput.log".

--- stderr
CMake Warning (dev) at CMakeLists.txt:8 (project):
  Policy CMP0048 is not set: project() command manages VERSION variables.
  Run "cmake --help-policy CMP0048" for policy details.  Use the cmake_policy
  command to set the policy and suppress this warning.

  The following variable(s) would be set to empty:

    CMAKE_PROJECT_VERSION
    CMAKE_PROJECT_VERSION_MAJOR
    CMAKE_PROJECT_VERSION_MINOR
[...]

Issue #383, opened 2019-08-16
`nannou_gui` - A native, more efficient, data-driven approach to UI compatible with future GUI editor plans.


mitchmindtree:
This issue is an actionable follow-up to the conclusions drawn in #2. Here I'll track my progress and plans, updating this top-level comment as I go.

Much of the work will draw from my work on conrod and the lessons learned during that process. That said, `nannou_gui` will differ in some fundamental ways:

- `nannou_gui` will remain tightly associated with `nannou`, leveraging its existing color, geometry, mesh and windowing abstractions. This will solve some annoying, existing type compatibility issues that currently exist in nannou's `Ui`, such as the need to use different color types between the UI and Draw APIs.
- `nannou_gui` will be based on a "retained-mode" API rather than an "immediate-mode" API. There are a few reasons for this:
    - **Performance**. The existing immediate mode API is more than efficient enough for projects requiring small, non-complex GUIs, but can become quite taxing on CPU in large, sophisticated applications (see [spatial_audio_server](https://github.com/museumsvictoria/spatial_audio_server)). This is due to the requirement for updating all visible widgets once every `update`, regardless of whether or not they have changed. That said, the ergonomics provided by immediate mode GUIs are highly beneficial. It should be possible to implement an immediate-mode API over the retained-mode API with an approach involving a lot of caching, which is largely how any slightly-efficient immediate mode GUI works internally anyway.
    - **GUI Editor**. `nannou_gui` is designed with GUI editor usage in mind - that is, the ability to design and develop efficient GUIs via a GUI. This requires the ability to treat widgets and their events as serializable data, an approach that is not currently possible with the existing immediate mode API in an efficient manner.

## Roadmap

- [ ] Create repository
- [ ] Events - UI event types and functions for interpreting them from raw windowing events.
- [ ] Graph - The data structure describing widget ownership and layout.
- [ ] Widget - a trait implemented for all widget types. Either this trait, or a closely related `SerdeWidget` trait should require `Deserialize` and `Serialize` and allow for trait object serialization via the `typetag` crate. While designing this interface, it will be important to consider that it should be possible to create widgets and their implementations entirely from scratch at run-time.
- [ ] Theme - a flexible data-structure allowing for the association of unique styling defaults per widget.
- [ ] Primitive Widgets - a small collection of widgets from which all other widgets may be composed:
    - [ ] Mesh - indexable triangulation consisting of a collection of vertex attribute channels. E.g. position, texture coordinates, colour, normals.
[...]

PR #615, opened 2020-05-29
`nannou_ui` crate WIP

mitchmindtree:
Currently a very, very early WIP of #383 and #430  It may not compile, and is likely to change often as I continue to experiment and try out different things. I thought I'd open a PR so it's easier for folks to keep track of progress and make suggestions. I'll update this comment once I have something a little more solid.


Type1J:
This is a great idea. I've been wanting to do [something like ScaleForm is able to do as easily as they do it](https://www.youtube.com/watch?v=zKDuzVbi50Q) for a while, now. I usually have to go to great lengths to get it to work due to reimplementing it every time.

PR #617, opened 2020-06-12
set minimum interval for request_redraw calls

mweitzel:
This PR adds a minimum time duration between `window.window.request_redraw()` attempts.

On OS X, when a window is entirely covered by other windows, [window.window.request_redraw](https://docs.rs/winit/0.20.0-alpha2/winit/window/struct.Window.html#method.request_redraw) makes a blocking call but does not actually perform the redraw. This results in apply_update/request_redraw being called every 3-10µs, which has two side effects:

- rust fully consumes a single core of the cpu
- the gpu is hammered with `request_redraw`s, causing whichever program is in the foreground to experience severe visual lag

This PR prevents superfluous redraw requests. It does _not_ attempt to short-circuit any `apply_update` calls past that. I'm not familiar enough with this project to know the side effects that would have.

### DIFF
--- a/nannou/src/app.rs
+++ b/nannou/src/app.rs
@@ -24,6 +24,7 @@ use std::collections::HashMap;
 use std::path::PathBuf;
 use std::sync::atomic::{self, AtomicBool};
 use std::sync::Arc;
+use std::time;
 use std::time::{Duration, Instant};
 use winit;
 use winit::event_loop::ControlFlow;
@@ -1062,6 +1063,7 @@ fn run_loop<M, E>(
                         .expect("missing swap chain");
                     let nth_frame = window.frame_count;
                     window.frame_count += 1;
+                    window.tracked_state.last_redraw_requested = Instant::now();
                     (swap_chain, nth_frame)
                 };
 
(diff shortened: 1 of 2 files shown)


mitchmindtree:
Thanks so much for the PR @mweitzel! I like where this is going.

We did have something very similar when using `vulkano` for graphics, however I never got around to investigating how to go about this following the switch to `wgpu-rs`.

There are a couple of things I think we should look into before landing this:

1. It would be nice if the user could specify the custom minimum interval. We used to have this as a parameter on the necessary `LoopMode` variants, but maybe it's easier to store this as an `Option<Duration>` within the `App` struct itself?
2. For apps running in the [`Wait` loop mode](https://docs.rs/nannou/0.14.1/nannou/app/enum.LoopMode.html#variant.Wait), I think this PR might result in a final update not being presented until some kind of user interaction occurs again. E.g. if the window is being resized and a request for redraw is omitted right before resize completes, the graphics may not appear to correctly fill the entire window until some other user interaction occurs in order to trigger a new update and in turn a new `request_redraw`.
   We could possibly work around this by 1. keeping track of whether or not each window requires another redraw request after the minimum interval has been exceeded and 2. generating some timed event that fires once the minimum interval has been exceeded so that we can make a final redraw request before the event loop goes back to "sleep" awaiting user input.

Let me know your thoughts.

Issue #619, opened 2020-06-21
Update Guide example to use nannou 0.14


oseph:
The nannou depenceny version in step 3 of the [Create a Project](https://guide.nannou.cc/getting_started/create_a_project.html) example should be updated to 0.14. 

With the current 0.12 version nannou throws a compile error expecting a reference to the Frame object.

mitchmindtree:
Thanks for the issue @oseph! Is there any chance you'd be interested in opening a PR for this?

oseph:
Sure! [Here's the PR.](https://github.com/nannou-org/nannou/pull/620)

mitchmindtree:
Closed via #620.

PR #620, opened 2020-06-24
Update `Create a Project` example to use nannou version 0.14

oseph:
Now this bare minimum example should compile and run without error ;-)


oseph:
#619

mitchmindtree:
Perfect, thanks @oseph!

Issue #596, opened 2020-05-10
Fix orientation of teapot in `wgpu_instancing.rs` after update to 0.5


mitchmindtree:
I just realised that I forgot to correct the orientation of the teapot in this example to account for wgpu's switch to having the y axis increase upwards (rather than downwards). Previously the teapot examples all used a view matrix with an inverted axis in order to account for the downwards y axis, as the original teapot vertices were intended for an upward y axis. This has been correct in the `wgpu_teapot*` examples in #594, but I forgot to correct it in `wgpu_instancing.rs`. This is a reminder to myself (or anyone) to fix it.

nidnogg:
Hey there! Shouldn't this be closed, considering the solution has been merged?

PR #622, opened 2020-07-01
Fix orientation of the teapot y axis in wgpu_instancing example

stephlow:
As noted in #596, this fixes the y axis for the wgpu_instancing example. Let me know if I missed something and I'll pick it up!


mitchmindtree:
Thanks @stephlow!

PR #623, opened 2020-07-02
Fix broken urls

sbauer322:
Just a quick change so the url works correctly in the guide site.


JoshuaBatty:
Great thanks for this @sbauer322

PR #625, opened 2020-07-11
Fix wrong comment on examples/draw/draw_mesh.rs

jmg-duarte:
The comment stated "Clear the background to blue" where the function clearly painted the background black.


mitchmindtree:
Thanks!

Issue #599, opened 2020-05-11
`draw.to_frame` panics if called without any graphics to draw


mitchmindtree:
I made some improvements to the way the `draw::Renderer` generates commands for switching between bind groups, render pipelines and drawing recently, but I think now it optimises out setting the pipeline and bind group for the render pass altogether if there's no graphics to be drawn (which is bad). It should be easy enough to add a special case that checks for this.

PR #636, opened 2020-07-28
Fix empty mesh panic

hypopyon:
Renderer::encode_render_pass no longer panics when given an empty Draw. Previously, it would attempt to create empty buffers, which is asserted to be false in wgpu::Device::create_buffer_mapped.


hypopyon:
#599

mitchmindtree:
Thanks so much @hypopyon!

PR #641, opened 2020-08-04
Fix generative design overflows

sbauer322:
p_2_1_1_03 was updated to better mirror the behavior for keys 3 and 4 found in https://editor.p5js.org/generative-design/sketches/P_2_1_1_03 - basically the transparency is calculated per line rather than one transparency for all lines. If you'd rather this be in a separate PR then let me know.

The rest of the changes for the three files were related to avoiding the overflow errors when not running with `--release`


mitchmindtree:
Thanks so much @sbauer322!

PR #642, opened 2020-08-22
Correct small mistake in coordinate_degrees functions documentation

TrickyRicky26:
Correct documentation for {x,y,z}_degrees functions to say "in degrees" instead of "in radians", which is incorrect.


mitchmindtree:
Thanks @TrickyRicky26!

Issue #435, opened 2019-12-11
Add audio input stream examples


mitchmindtree:
To begin, it might be nice to port [CPAL's `record_wav.rs`](https://github.com/RustAudio/cpal/blob/master/examples/record_wav.rs) example.

Following this, CPAL's `feedback.rs` example might also be useful for demonstrating how to run an input and output stream simultaneously and feed audio from one to the other. This might be especially useful seeing as we don't yet have a duplex audio stream API just yet.

wjakobczyk:
I guess there not much left to be done here, right?

PR #648, opened 2020-08-27
Port CPAL's record_wav example to nannou

zehreken:
The example shows how to create an input stream, record audio and write it to disk in .wav format. Because of nannou's interactive nature, I couldn't help but add a record indicator and basic controls to play and pause the stream. Related to issue #435


mitchmindtree:
Really nice, thanks heaps @zehreken :)

Issue #647, opened 2020-08-27
Update wgpu to 0.6


yutannihilation:
I tried to create a pull request for this, but it seems too difficult to me... So, let me just file an issue.(My failed attempt is here: <https://github.com/nannou-org/nannou/compare/master...yutannihilation:poc/wgpu-0.6.0_tmp?expand=1>)

Some notes:

* conrod needs to be updated. I created a pull request for this: https://github.com/PistonDevelopers/conrod/pull/1385#pullrequestreview-476458514
* Since wgpu implemented new `map_async()` in https://github.com/gfx-rs/wgpu-rs/pull/344 and https://github.com/gfx-rs/wgpu/pull/675, there's no `BufferReadMapping` anymore. Maybe `image.rs` needs some refactoring?
* `array_layer` is moved to other place in https://github.com/gfx-rs/wgpu/pull/666. I couldn't figure out how this is passed around in nannou, but I hope this shouldn't be very hard to fix.
* I needed to add tons of explicit lifetimes as more structs requires it. Not sure if this is just my skill about lifetime is very poor...

mitchmindtree:
Thanks for diving in @yutannihilation and for sharing what you learned!

> I needed to add tons of explicit lifetimes as more structs requires it.

It looks like you're not the only one to have run into this - I added a comment about addressing this on the active PR [here](https://github.com/nannou-org/nannou/pull/655#discussion_r499248832) in case you're interested.

yutannihilation:
Thanks! Yeah, your comment on #655 is really helpful :)

PR #655, opened 2020-09-25
Wgpu 0.6

alekspickle:
Here is my attempt to migrate wgpu from 0.5 to 0.6. 
Did it the most straightforward way I can come up with so feel free to disregard it. 

But it would be cool if someone point me out on why it is stuck compiling nannou...


nannou/src/wgpu/texture/mod.rs
```diff
@@ -854,6 +854,20 @@ pub fn format_to_component_type(format: wgpu::TextureFormat) -> wgpu::TextureCom
...
+        | wgpu::TextureFormat::Bc3RgbaUnorm
+        | wgpu::TextureFormat::Bc3RgbaUnormSrgb
+        | wgpu::TextureFormat::Bc4RUnorm
+        | wgpu::TextureFormat::Bc4RSnorm
+        | wgpu::TextureFormat::Bc5RgUnorm
+        | wgpu::TextureFormat::Bc5RgSnorm
+        | wgpu::TextureFormat::Bc6hRgbSfloat
+        | wgpu::TextureFormat::Bc6hRgbUfloat
+        | wgpu::TextureFormat::Bc7RgbaUnorm
+        | wgpu::TextureFormat::Bc7RgbaUnormSrgb => wgpu::TextureComponentType::Float,
```
yutannihilation, line 871: Now that https://github.com/gfx-rs/wgpu/pull/628 is merged, `format_to_component_type()` is no longer needed (we just can use `into()`).

nannou/src/draw/renderer/mod.rs
```diff
@@ -413,7 +417,7 @@ impl Renderer {
         let uniforms = create_uniforms(output_attachment_size, output_scale_factor);
         let uniforms_bytes = uniforms_as_bytes(&uniforms);
         let usage = wgpu::BufferUsage::UNIFORM | wgpu::BufferUsage::COPY_DST;
-        let uniform_buffer = device.create_buffer_with_data(uniforms_bytes, usage);
+        let uniform_buffer = device.create_buffer(uniforms_bytes, usage);
```
yutannihilation, line 420: The equivalent in wgpu 0.6 to `create_buffer_with_data()` is [`create_buffer_init()`](https://docs.rs/wgpu/0.6.0/wgpu/util/trait.DeviceExt.html#tymethod.create_buffer_init).

nannou/src/draw/renderer/mod.rs
```diff
@@ -395,7 +399,7 @@ impl Renderer {
             .usage(wgpu::TextureUsage::SAMPLED | wgpu::TextureUsage::COPY_DST)
             .format(Self::GLYPH_CACHE_TEXTURE_FORMAT)
             .build(device);
-        let glyph_cache_texture_view = glyph_cache_texture.create_default_view();
+        let glyph_cache_texture_view = glyph_cache_texture.view().build();
```
yutannihilation, line 402: The equivalent in wgpu 0.6 to `create_default_view()` is `create_view(&wgpu::TextureViewDescriptor::default())`.

yutannihilation (commented):
Cool. I tried this and failed (just because my Rust skill is very poor)... Let me leave some quick comments that I've learned from the failure.

(The full diff of my attempt and the issues I couldn't handle are listed on #647)

mitchmindtree:
Amazing, thanks for such an epic PR!

Unfortunately I'm busy with a contract at the moment, but will take a proper look as soon as I can!

nannou/src/wgpu/texture/mod.rs
alekspickle: [8da1877 ](https://github.com/nannou-org/nannou/pull/655/commits/8da1877e80e364f04ad7d0dc6c899f39847c6354)

nannou/src/draw/renderer/mod.rs
alekspickle: [8ed73cb ](https://github.com/nannou-org/nannou/pull/655/commits/8ed73cb3fc891e1901080edc115f37369a4c1d0d)

nannou/src/draw/renderer/mod.rs
alekspickle: [1cf9fce](https://github.com/nannou-org/nannou/pull/655/commits/1cf9fce293cffbfdd92d50e18db80f674dfcc4ae)

alekspickle:
- removed **format_to_component**
- adjust buffers init and default view creation to latest wgpu

nannou/src/wgpu/texture/mod.rs
```diff
@@ -35,9 +35,9 @@ pub struct Texture {
 /// be of the whole texture, but it might also be of some sub-section of the texture. When an API
 /// provides
 #[derive(Debug)]
-pub struct TextureView {
+pub struct TextureView<'t> {
     handle: Arc<TextureViewHandle>,
-    descriptor: wgpu::TextureViewDescriptor,
+    descriptor: wgpu::TextureViewDescriptor<'t>,
```
mitchmindtree, line 40: Hey @alekspickle, it looks like you might be running into some issues with Rust's lifetimes, and my suspicion is that this field here is the source of the unexpected complexity :) 

The lifetime parameter on the `wgpu::TextureViewDescriptor` is associated with the lifetime of its `label` field. The purpose of this `label` field is to make it easier to track `TextureView`s when debugging or using some inspector like renderdoc. By remaining generic over the lifetime, wgpu allows users to use a label that is a reference to some static slice of memory (e.g. string literals like "foo", `const` or `static` `str`s) or dynamically allocated one (e.g. a reference to a `String`).

Throughout nannou, the only descriptor labels we use are string literals (e.g. "frame-texture-vew") as these are almost always enough to quickly identify the resource in question when using a graphics debugger / inspector. As a result, our `wgpu::TextureViewDescriptor`'s always have `'static` lifetimes. In case you're unfamiliar, `'static` lifetimes are special in that they indicate a lifetime is valid for the entire duration the program will run.

We should be able to get away without `TextureView` having a lifetime parameter, and instead declaring the `descriptor` field like this:

```rust
pub struct TextureView {
    handle: Arc<TextureViewHandle>,
    descriptor: wgpu::TextureViewDescriptor<'static>, // notice the 'static lifetime here
    texture_extent: wgpu::Extent3d,
    texture_id: TextureId,
}
```

This should hopefully allow you to avoid this problem where the lifetime parameters seem to propagate up through the whole program and it becomes impossible to make the borrow checker happy.

I hope this makes sense / helps a bit!

alekspickle: Indeed, that could simplify things. 
But I guess the reason is me understanding wgpu/nannou code intricacies enough.

kazimuth:
I was interested in this for my own reasons (wanted to try some new wgpu-0.6 stuff in my sketches :), spent some time poking through this PR.

@alekspickle:

> But it would be cool if someone point me out on why it is stuck compiling nannou...

This happens on my machine too. It looks like `rustc` is hanging! Somewhere in the texture management code. It's weird, I've never seen that happen before. I think it's because type resolution gets confused inside one of the async methods? Definitely a compiler bug of some kind.

I made some changes [here](https://github.com/alekspickle/nannou/compare/wgpu-0.6...kazimuth:wgpu-0.6-redux) to fix that. I also went and changed some of the lifetimes assuming that the wgpu `Descriptor` objects would have `'static` labels, which simplifies things a lot. There's a fresh wall of errors to dig through now, which is exciting :) I'm thinking of poking at that some more this weekend, but if you'd rather be in charge let me know @alekspickle .

Also, it does appear that there are some architectural changes in 0.6 that need to be handled.
- [`create_surface`](https://docs.rs/wgpu/0.6.0/wgpu/struct.Instance.html#method.create_surface) is now an unsafe function on `Instance`. So I think we'll need to take the `Instance` created [here](https://github.com/nannou-org/nannou/pull/655/files#diff-74fd73b81728e081f92ed6e8992b6d10R131) and store it... somewhere, not sure where that should go. Maybe just in `App`, or as part of the device map? @mitchmindtree , do you have thoughts?
- `BufferReadMapping` and `BufferWriteMapping` have been replaced with  [`BufferSlice`](https://docs.rs/wgpu/0.6.0/wgpu/struct.BufferSlice.html), which carries a lifetime. That means that `ImageReadMapping` and `Rgba8ReadMapping` will need to grow lifetimes as well. Not sure if that'll have any larger impact.

mitchmindtree:
> This happens on my machine too

@kazimuth Likewise! I've never seen it before either - I ended up having a crack at this from the ground up, see #660. 

> Maybe just in App, or as part of the device map?

Yeah these were my thoughts too, I think it might best live alongside the `adapters` (if they're still needed that is, I haven't checked) in the `App`.

I'm not yet sure when I'll get a chance to continue on #660, but feel free to take what you like from it if or fork it if you're hacking on your own branch :) 

> BufferReadMapping and BufferWriteMapping have been replaced with BufferSlice, which carries a lifetime. That means that ImageReadMapping and Rgba8ReadMapping will need to grow lifetimes as well. Not sure if that'll have any larger impact.

Yeah I imagine this might be the trickier change. I haven't yet taken the time to see how much those lifetimes propagating through types will affect things (I guess this will depend a lot on we store them or users are encouraged to store them), it might be doable but it's also possible we'll need to approach this in a different way. The [wgpu users channel on matrix](https://matrix.to/#/!XFRnMvAfptAHthwBCx:matrix.org?via=matrix.org&via=kyju.org&via=matrix.parity.io) might be a great place to ask.

alekspickle:
@kazimuth a lot of errors after you fix seems like those I resolved before. I dont understand where they come from though, because it was eliminanted by using **'static** lifetime.

kazimuth:
@alekspickle I think that's because rustc was dying before it got to those errors, so you wouldn't get the error messages. I get them too, will hold off on trying to fix if you're working on it. (You're welcome to merge my changes into this PR if you want, no pressure though.)

@mitchmindtree Doing a more in-depth read of the code, lifetimes will need to be added to:
- `BufferBytes`
- `BufferImage`
- `Snapshot`
- `Rgba8ReadMapping`
- `Rgba8AsyncMappedImageBuffer`
Specifically, they all need to have the lifetime of the `Buffer` that is being read. That `Buffer` is ultimately created in `Capturer::capture()`. So unfortunately, there isn't really a way to pass them back to the user while using that lifetime. You'd need to either:
1. use one of those monstrosity crates that allow self-referential structs and return them as a bundle
2. make the api two-step: have the user call one thing to hold the `Buffer`, and then another call receives that as a reference and returns a `Snapshot` that borrows it... asynchronously.
3. switch the API to only use callbacks that receive a `Snapshot`.

Of these options, I think the third is probably the simplest, both from an implementation and use perspective.

alekspickle:
@kazimuth Yeah, you've got the point. feel free to poke it yourself, I stumbled at least 2 things I dont know how to solve, which require more of both nannou and wgpu-rs experience.

mitchmindtree:
@kazimuth I wonder if we can offer a combination of your 2nd and 3rd options? E.g. having two separate methods for allowing the user to produce either a future which they can handle somehow themselves, or pass a callback which nannou can call internally once the future completes? I haven't yet done enough digging into the new async read/write API in wgpu to know how feasible this would be.

kazimuth:
@mitchmindtree it's probably doable.

Other question: the `array_layer_count` attribute was removed from `TextureDescriptor` in 0.6, but it was kept on `TextureViewDescriptor`. Unfortunately the texture builder code seems to use that attribute a lot. I'm not sure what to substitute to keep the logic the same. Do you have any idea what's going on there? I'm not all that familiar with array textures. I'd read the spec, but... [well](https://gpuweb.github.io/gpuweb/#array-layer)...

kazimuth:
Update: apparently `array_layer_count` now only exists on `TextureViews`. A `Texture` may be 1d, 2d, or 3d, but only a texture view can be an array, cubemap, or cube array. I think I'll update the API to reflect this.

kazimuth:
New work: https://github.com/alekspickle/nannou/compare/wgpu-0.6...kazimuth:wgpu-0.6-3

Don't merge it yet, I still need to make the changes with Buffer borrows.

kazimuth:
Update: all compile errors fixed on my branch!

...Now we just have to address the runtime errors :sweat_smile: 

Things that need to be addressed:
- All of the examples result in blank screens. I'll do some poking around in RenderDoc, bet there's a new flag somewhere with the wrong value.
- There's now a [256-byte alignment requirement](https://docs.rs/wgpu/0.6.0/wgpu/constant.COPY_BYTES_PER_ROW_ALIGNMENT.html) for texture-to-buffer and buffer-to-texture copies, which breaks Capturer. We might be able to fix this by just... padding and reshaping the textures? Not sure.
- In general, the tires should probably get kicked a little more. Going through and running different examples is probably a good idea.

PR #660, opened 2020-10-09
Updating to wgpu 0.6 - follow-up attempt to #655

mitchmindtree:
This is a WIP follow-up to #655, attempting to update to wgpu 0.6.

I was running into an issue where rustc would hang indefinitely, and there were quite a few lifetime issues I wanted to address so thought I would have a crack at this from scratch.

TODO:

- [ ] Store `Instance` somewhere at app-level, maybe with device map.
- [ ] Provide builder methods for specifying 'static labels.
- [ ] Solve capturer Snapshots async changes.
- [ ] Review `wgpu::create_pipeline_layout` - can be replaced with a builder?
- [ ] Review IntoPipelineLayoutDescriptor.

cc @alekspickle @yutannihilation @kazimuth

### DIFF
--- a/nannou/Cargo.toml
+++ b/nannou/Cargo.toml
@@ -15,9 +15,9 @@ default = ["notosans"]
 
 [dependencies]
 cgmath = { version = "0.17", features = ["serde"] }
-conrod_core = "0.70"
-conrod_wgpu = "0.70"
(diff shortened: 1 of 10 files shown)


mitchmindtree:
No longer necessary as of #665.

PR #662, opened 2020-10-10
Update nannou_audio to use cpal 0.12 and replace sample crate with dasp_sample 

JoshuaBatty:
This is nowhere near ready but i'm putting this up so people can follow along and leave comments. 

Quite a lot has changed between cpal 0.11 and cpal 0.12. Specifically, we can now request a BufferSize (great for low latency applications, which I assume is most nannou apps!) and the removal of the existing EventLoop. 

The `sample` crate has also been deprecated in favour of `dasp`. Instead of importing the whole of `dasp` I've only included the `dasp_sample` crate as it has everything that we need.

TODO:

- [ ] Remove the cpal event loop

### DIFF
--- a/nannou_audio/Cargo.toml
+++ b/nannou_audio/Cargo.toml
@@ -11,9 +11,9 @@ homepage = "https://nannou.cc"
 edition = "2018"
 
 [dependencies]
-cpal = "0.11"
+cpal = "0.12"
 failure = "0.1"
-sample = "0.10"
+dasp_sample = "0.11.0"
 
 [features]
 asio = ["cpal/asio"]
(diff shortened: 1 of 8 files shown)


mitchmindtree:
After Matrix chat:

- Move stream creation methods to `Device`.
- Remove `EventLoop` entirely - shouldn't be necessary anymore.
- Should be able to remove `LoopContext` too.
- Remove `StreamId`.

Hopefully from there it should all fall into place, I think CPAL is much closer to nannou_audio API-wise these days than it used to be.

JoshuaBatty:
closing this as @mitchmindtree implemented the rest in #670

PR #667, opened 2020-10-27
Refactor: use shorthand fields

jmg-duarte:
The title says it all.

It was automated using [comby](https://comby.dev/) and this [pattern](https://catalog.comby.dev/#redundant-field-names-single).

As a suggestion, maybe this could be automated with GitHub Actions


mitchmindtree:
Thanks!

Issue #315, opened 2019-05-22
missing (real) MIDI support


tpltnt:
Hi there,

the readme claims MIDI support. The only related code I could find is `OscMidiMessage` which handles MDI over OSC, but does not involve local MIDI devices. Either there should be support for MIDI devices which are attached locally or the claim should be removed (or at least made be less misleading.

Cheers,
tpltnt

mitchmindtree:
Thanks for the issue!

I totally see where you are coming from - sorry for the confusion! The intention has been to add an example that demonstrates how easy it is to use MIDI via [`midir`](https://crates.io/crates/midir) in a nannou application (I've done so in the past and I believe @freesig has also, though neither projects are open source at the moment unfortunately). I think first writing some MIDI examples will also lead us toward any higher-level abstraction the nannou-org might be able to provide on top in a `nannou-midi` crate or something along those lines. It will also enable us to constantly test interop between the two crates via CI.

Perhaps it's worth separating out the features that nannou implements directly from those that a creative coder might be interested in that can be easily used via an external crate for clarity? Others that come to mind are DMX via sacn, serial via serial and TCP/UDP via std (each of these link to their respective crates in the README). This way we can still demonstrate to folks coming from environments like OF or Processing that MIDI is just as accessible in nannou as it is via an addon in one of those libraries, while still distinguishing between the support that we provide from within the organisation and those common creative coding protocol crates that work well alongside it.

We're currently working on [the guide](https://github.com/nannou-org/guide) with which we're planning on replacing the vast majority of the README contents soon, so this is a timely issue :+1:

tpltnt:
I would like to see a working example since I am currently grinding my teeth on using `midir` and nannou to read MIDI messages. I can see some data, but failed to track/integrate it into the (global) model.

lukors:
I would also like to have some example of using MIDI in Nannou.

You should absolutely clarify what is actually provided by Nannou and what is simply another crate that can be used with Nannou. I saw MIDI in the list and got quite confused when I found no mention of MIDI in the source code and no tutorials on using MIDI and no examples.

[...]

PR #668, opened 2020-10-28
Guide: Unchecked MIDI support

lukors:
Also added explanation that midir can be used for the same purpose.

Related to #315


mitchmindtree:
Thanks!

PR #675, opened 2020-11-13
Remove incorrect comment

lpil:
Hello! That's for this delightful project!

I think this was accidentally copied from the other example, so I've removed it.

Cheers,
Louis


mitchmindtree:
Thanks so much for the kind words and patch!

PR #687, opened 2020-12-09
update Getting Started "vulkan-utils" reference for Debian Buster and other guide changes

seeplusplus:
Nothing huge, just came across this when setting up to use nannou on Debian Buster. The vulkan-utils package is transitional now. I'm unsure if this affects any other distros.


mitchmindtree:
Thanks for the PR!

PR #692, opened 2020-12-19
Fix non-uniform scaling

mgunyho:
Hello, I'm just starting out with nannou and it seems amazing so far!

I noticed that the non-uniform scaling functions `scale_x`, `scale_y` and `scale_z` scale all other dimensions to zero. I suppose this is unintentional. Or am I misunderstanding something?


mitchmindtree:
Ahh good spot! These methods are quite new and yes, this is certainly unintentional. Thanks a lot for taking the time to submit a PR :)

PR #696, opened 2021-01-06
Fix osc version in the guide

narisada014:
Hi, I've been following the guide and doing the tutorials. 
When I checked the latest version of nannou_osc, it was different from the one in the guide.
スクリーンショット 2021-01-07 1 09 40


mitchmindtree:
Thanks for the PR!

Issue #475, opened 2020-03-06
Possibility of supporting webassembly target for hosting designs on websites?


jasikpark:
I'm sure this is far in the future roadmap, but I think it'd be really cool to be able to compile a sketch to webassembly to be then executed as a page on a website.

I haven't really looked at this project yet 😅 so y'all might already support wasm? But there's definitely nothing in the guide, and I'm sure there's some particular kinks to work out with publishing on wasm, so I'll be interested to see how that plays out!!

This looks awesome btw!

mitchmindtree:
For sure, we have been thinking a lot on this! This was one of the major motivations for #452. There is work underway in both the audio and graphics backends to make it possible. RustAudio/cpal#372 and gfx-rs/wgpu-rs#101 are the two to watch. Then it's a matter of tracking [WebGPU adoption in browsers](https://github.com/gpuweb/gpuweb/wiki/Implementation-Status) and waiting for them to announce support in their stable releases.

Type1J:
Duplicate of #7?

Type1J:
With https://github.com/gfx-rs/wgpu-rs/issues/101 closed, what's left?

mitchmindtree:
The next steps for getting `nannou` running in the browser will be:

- Update to wgpu 0.5.
- Follow the steps outlined in [the recent wgpu-rs blogpost](https://gfx-rs.github.io/2020/04/21/wgpu-web.html) - specifically, the "Getting started with wgpu-rs on the web" section. This will likely unvail a few other small web-target-incompatible corners of nannou (e.g. things like file saving / loading) however it should be easy enough to feature-gate these so that they are only available on non-web targets, or to change them to be web-target-friendly.

For `nannou_audio`, RustAudio/cpal#372 will still be necessary. As for graphics however, it seems we're getting closer!

Type1J:
How close are we to web, now?

AJTJ:
Indeed, how close are we? I would love to use Nannou to make interactive designs with WASM. Is there some way this process can be supported?

Type1J:
I'm hoping that it will be, soon. D3 is nice, but mixing in some 3D is easier with Nannou. This may depend on the WebGPU standard (or draft) being supported in browsers due to web

Woyten:
I tried hard to port [microwave](https://crates.io/crates/microwave) to the browser until I came to the conclusion that it is not possible at the moment and probably for a longer period of time.

Steps that can be done right now:
- Fix dependency setup
[...]

PR #702, opened 2021-01-19
Use instant crate for WASM compatibility

Woyten:
Use `instant::Instant` instead of `std::time::Instant`.

`instant::Instant` is a type alias for `std::time::Instant` on non-WASM environments. On WASM environments, however, `Instant` is replaced by a browser-friendly implementation which doesn't crash when `Instant::now` is called (which would be the case with `std::time::Instant`).

Relates to #475


mitchmindtree:
Thanks @Woyten, seems a good place to start :)

PR #709, opened 2021-02-16
Fix warnings emitted when running draw example

babbageclunk:
They confused me a bit when building for the first time (much slower) because I didn't realise they were warnings until the app window popped up, so I figured it would be worth saving someone else that confusion. :smiley: 

The bulk of them are converting deprecated `to_rgba` calls to `to_rgba8`, also adding a missing debug value to `unwrap_or_panic`.

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -14,7 +14,7 @@ homepage = "https://github.com/nannou-org/nannou"
 edition = "2018"
 
 [dev-dependencies]
-audrey = "0.2"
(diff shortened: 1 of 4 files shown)


babbageclunk:
Of course I needed to include the audrey dependency bump for it to pass CI, whoops!

babbageclunk:
Repushed the dependency bump commit - the last CI run failed downloading crates from cargo.

mitchmindtree:
Thanks for the PR!

PR #712, opened 2021-02-25
Add a method to quit the application

### LINKED ISSUE #377: Exiting the application.
I am trying to find a way to trigger a graceful exit from within nannou (something that runs the specified exit function before quitting for example) 

Is there a way to implement this right now?

sumibi-yakitori:
Closes #377


mitchmindtree:
This is a really clever-yet-simple solution! Thanks :)

PR #715, opened 2021-03-01
Make nannou package compile in WASM environment

Woyten:
Hi. This is a minimal example showing which changes will allow to compile the nannou module against the `wasm32-unknown-unknown` target.

Disclaimer: I am not saying that:
- Nannou Apps will run in a Browser after these changes. There are still some things to do.
- This PR should be merged. But it could! :wink:

This PR relates to #475.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,2 @@
+[target.wasm32-unknown-unknown]
+rustflags = ["--cfg", "web_sys_unstable_apis"]
--- a/.github/workflows/nannou.yml
+++ b/.github/workflows/nannou.yml
@@ -113,11 +113,17 @@ jobs:
         profile: minimal
         toolchain: stable
         override: true
+        target: wasm32-unknown-unknown
     - name: Run check
       uses: actions-rs/cargo@v1
       with:
         command: check
         args: --examples --verbose
+    - name: Check nannou package builds against wasm32-unknown-unknown
+      uses: actions-rs/cargo@v1
+      with:
+        command: build
+        args: -p nannou --features wasm-experimental --target wasm32-unknown-unknown
 
   cargo-publish:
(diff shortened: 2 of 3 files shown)


mitchmindtree:
Nice to see a start on this!

One thing that might be worth adding while landing wasm support would be one new [CI job](https://github.com/nannou-org/nannou/blob/master/.github/workflows/nannou.yml) that tries to build one or more of the examples for the wasm target, or something along these lines. Normally I would say CI testing of specific platforms is best left for the cross platform crates like winit and cpal. However in our case I think it might be easy for us to accidentally break wasm compatibility otherwise, particularly as some library APIs just aren't compatible with wasm (like those wgpu items you had to gate in this PR) and it's not always obvious what those are.

Woyten:
Sorry I haven't responded yet. I will rebase this PR and see how to adapt the CI job once #737 is merged.

Woyten:
@mitchmindtree I upgraded this PR to the newest Nannou version and added a CI check.

Woyten:
@mitchmindtree Currently, it's not possible to compile the whole Nannou tree to WASM. `nannou_timeline` depends on `envelope` which depends on `num` which depends on `rustc-serialize`. `rustc-serialize` will not receive WASM support (https://github.com/rust-lang-deprecated/rustc-serialize/pull/190#issuecomment-378057459). This means, a new `envelope` version without the implicit `rustc-serialize` dependency is needed.

Woyten:
Other problematic libraries that need to be removed or hidden behind a feature switch are `socket2` and `home`.

Woyten:
@mitchmindtree The PR has been updated to the newest Nannou release. I propose to merge the PR s.t. the main `nannou` package doesn't lose WASM compatibility in the future.

Issue #690, opened 2020-12-13
Error when resizing window completely


LowSpecCorgi:
When resizing the window completely, in a bare bones program, I am met with the following error (with bactrace):
<details> 
  <summary>Error, condensed into summary tags, to make this issue less verbose</summary>

```rust
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `4`,
 right: `1`', /home/basilicous/.cargo/registry/src/github.com-1ecc6299db9ec823/wgpu-core-0.5.6/src/conv.rs:459:13
stack backtrace:
   0: rust_begin_unwind
             at /rustc/7eac88abb2e57e752f3302f02be5f3ce3d7adfb4/library/std/src/panicking.rs:483
   1: std::panicking::begin_panic_fmt
             at /rustc/7eac88abb2e57e752f3302f02be5f3ce3d7adfb4/library/std/src/panicking.rs:437
   2: wgpu_core::conv::map_texture_dimension_size
             at /home/basilicous/.cargo/registry/src/github.com-1ecc6299db9ec823/wgpu-core-0.5.6/src/conv.rs:459
   3: wgpu_core::device::Device<B>::create_texture
             at /home/basilicous/.cargo/registry/src/github.com-1ecc6299db9ec823/wgpu-core-0.5.6/src/device/mod.rs:365
   4: wgpu_core::device::<impl wgpu_core::hub::Global<G>>::device_create_texture
             at /home/basilicous/.cargo/registry/src/github.com-1ecc6299db9ec823/wgpu-core-0.5.6/src/device/mod.rs:677
   5: wgpu_device_create_texture
             at /home/basilicous/.cargo/registry/src/github.com-1ecc6299db9ec823/wgpu-native-0.5.1/src/device.rs:225
   6: wgpu::Device::create_texture
             at /home/basilicous/.cargo/registry/src/github.com-1ecc6299db9ec823/wgpu-0.5.2/src/lib.rs:979
   7: nannou::wgpu::texture::Builder::build
             at /home/basilicous/.cargo/registry/src/github.com-1ecc6299db9ec823/nannou-0.15.0/src/wgpu/texture/mod.rs:559
   8: nannou::frame::create_lin_srgba_msaa_texture
             at /home/basilicous/.cargo/registry/src/github.com-1ecc6299db9ec823/nannou-0.15.0/src/frame/mod.rs:333
   9: nannou::frame::create_intermediary_lin_srgba
[...]

Issue #651, opened 2020-08-30
Timeout when acquiring next swap chain texture


jpopesculian:
Hey! I love nannou, but last week suddenly nannou seems to crash whenever I try to run a program. I'm running

* OS: Ubuntu 20.04.1 LTS
* Processor: Intel® Core™ i7-10510U CPU @ 1.80GHz × 8
* Graphics: Mesa Intel® UHD Graphics (CML GT2)
* Kernel: 5.4.0
* Windowing: X11

I should have the most up to date mesa drivers. I found this issue https://github.com/gfx-rs/wgpu-rs/issues/164 but the examples for `winit`, `wgpu` and `ash` all run fine. I don't know how to investigate the issue any further, so any help would greatly be appreciated.

Here's the error when I do `RUST_BACKTRACE=full cargo run --example draw`:

```
WARNING: vallium/llvmpipe is not a conformant vulkan implementation, testing use only.
INTEL-MESA: warning: Performance support disabled, consider sysctl dev.i915.perf_stream_paranoid=0

thread 'main' panicked at 'failed to acquire next swapchain texture: TimeOut', /home/julian/Development/nannou/nannou/src/app.rs:1069:45
stack backtrace:
   0:     0x55cd9c16ea85 - backtrace::backtrace::libunwind::trace::h396c07d2071b43af
                               at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.46/src/backtrace/libunwind.rs:86
   1:     0x55cd9c16ea85 - backtrace::backtrace::trace_unsynchronized::h7aa0e4bb23d9c158
                               at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.46/src/backtrace/mod.rs:66
   2:     0x55cd9c16ea85 - std::sys_common::backtrace::_print_fmt::hd15ac5d4adcd355b
                               at src/libstd/sys_common/backtrace.rs:78
   3:     0x55cd9c16ea85 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::hec5354be8ccc3ecc
                               at src/libstd/sys_common/backtrace.rs:59
   4:     0x55cd9c19601c - core::fmt::write::h3d34909eeb4f225b
                               at src/libcore/fmt/mod.rs:1076
   5:     0x55cd9c16b623 - std::io::Write::write_fmt::h1da287b3de55ed16
                               at src/libstd/io/mod.rs:1537
[...]

Issue #491, opened 2020-03-24
Crash when resizing window quickly multiple times.


edap:
When resizing the window quickly on linux, the application crash.

thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: AllocationError(OutOfMemory(Device))', src/libcore/result.rs:1165:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.

Version 0.13.1

mitchmindtree:
Thanks for the report! I'm also noticing this on Linux when resizing for a while. It seems we have some textures or some other resource that isn't being cleaned up properly. I'm not sure if its us that should be releasing the resources of it its a leak in wgpu. Will have to do some investigating with `renderdoc`!

tpltnt:
Can confirm. Longer backtrace:
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: AllocationError(OutOfMemory(Device))', src/libcore/result.rs:1165:5
stack backtrace:
   0: backtrace::backtrace::libunwind::trace
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.40/src/backtrace/libunwind.rs:88
   1: backtrace::backtrace::trace_unsynchronized
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.40/src/backtrace/mod.rs:66
   2: std::sys_common::backtrace::_print_fmt
             at src/libstd/sys_common/backtrace.rs:77
   3: <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt
             at src/libstd/sys_common/backtrace.rs:61
   4: core::fmt::write
             at src/libcore/fmt/mod.rs:1028
   5: std::io::Write::write_fmt
             at src/libstd/io/mod.rs:1412
   6: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:65
   7: std::sys_common::backtrace::print
             at src/libstd/sys_common/backtrace.rs:50
   8: std::panicking::default_hook::{{closure}}
             at src/libstd/panicking.rs:188
   9: std::panicking::default_hook
             at src/libstd/panicking.rs:205
  10: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:464
  11: std::panicking::continue_panic_fmt
             at src/libstd/panicking.rs:373
  12: rust_begin_unwind
             at src/libstd/panicking.rs:302
[...]

PR #720, opened 2021-04-06
Fix various resize crashes, texture memory leak

### LINKED ISSUE #676: Memory Leak on Windows when running example?
When running example generative_designs/p_2_2_5_01.rs on Windows, if I resize the window the memory used on the GPU keeps going as the resize is in progress.

I just cloned the repository and ran the example like so -

```shell
C:\Users\dhruv\Documents\Repositories\nannou\(master)> cargo run --release --example p_2_2_5_01
    Finished release [optimized] target(s) in 0.66s
     Running `target\release\examples\p_2_2_5_01.exe`
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: AllocationError(OutOfMemory(Device))', C:\Users\dhruv\.cargo\registry\src\github.com-1ecc6299db9ec823\wgpu-core-0.5.6\src\device\mod.rs:417:14
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
error: process didn't exit successfully: `target\release\examples\p_2_2_5_01.exe` (exit code: 101)
```

If I resize slowly, the memory grows slowly, and vice versa. I know this by looking at the memory used graph in task manager.

danwilhelm:
Fixes #676, #690, #491, possibly #651 (only tested on Mac and Windows).

**Overview**
1. Fixes major texture memory leak. Solved by replacing `Arc::into_raw` with `Arc::as_ptr`.
2. Prevents crashes when the window is resized < 2x2:
    1. Enforces a minimum swap chain and window size.
    2. Adds additional error handling that recreates the swap chain if it gets lost or is invalid.

**Notes**
+ By default, this change disallows resizing windows below 2x2. However, any minimum size can still be chosen by the user when building the window. (It still will not crash when 0x0, due to the other fixes.)
+ In testing, determined the swap chain had to be at least 2x2 given a 4x MSAA. However, the [wgpu-rs msaa-line example](https://github.com/gfx-rs/wgpu-rs/tree/master/examples/msaa-line) allows a minimum size of 1x1. Have not investigated the discrepancy.
+ Rust 1.45+ is now required due to using `Arc::as_ptr`.

### DIFF
--- a/nannou/src/app.rs
+++ b/nannou/src/app.rs
@@ -26,6 +26,7 @@ use std::path::PathBuf;
 use std::sync::atomic::{self, AtomicBool};
 use std::sync::Arc;
 use std::time::Duration;
+use wgpu_upstream::SwapChain;
 use winit;
 use winit::event_loop::ControlFlow;
 
@@ -749,6 +750,27 @@ impl App {
         windows.keys().cloned().collect()
     }
 
+    /// Return the **SwapChain** for the given `window_id`, rebuilding if it doesn't exist.
+    ///
+    /// **Panics** if `window_id` does not exist, or if the swap chain is missing and cannot be rebuilt.
+    pub fn get_or_rebuild_swap_chain(&self, window_id: window::Id) -> SwapChain {
+        let mut windows = self.windows.borrow_mut();
+
+        let window = windows
+            .get_mut(&window_id)
+            .expect("no window for redraw request ID");
+
+        if window.swap_chain.swap_chain.is_none() {
+            window.rebuild_swap_chain(window.tracked_state.physical_size.into());
+        }
+
+        window
+            .swap_chain
+            .swap_chain
+            .take()
+            .expect("missing swap chain")
+    }
+
     /// Return the **Rect** for the currently focused window.
     ///
     /// The **Rect** coords are described in "points" (pixels divided by the hidpi factor).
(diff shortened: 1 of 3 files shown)


nannou/src/wgpu/texture/mod.rs
```diff
@@ -202,7 +202,7 @@ impl Texture {
     /// This is useful for distinguishing between two **Texture**s or for producing a hashable
     /// representation.
     pub fn id(&self) -> TextureId {
-        TextureId(Arc::into_raw(self.handle.clone()) as usize)
+        TextureId(Arc::as_ptr(&self.handle) as usize)
```
mitchmindtree, line 205: Agh great find! :tada: 

I'm rather embarrassed that this snuck through in the first place, very glad you were able to spot it :blush:

nannou/src/app.rs
```diff
@@ -749,6 +750,27 @@ impl App {
         windows.keys().cloned().collect()
     }
 
+    /// Return the **SwapChain** for the given `window_id`, rebuilding if it doesn't exist.
+    ///
+    /// **Panics** if `window_id` does not exist, or if the swap chain is missing and cannot be rebuilt.
+    pub fn get_or_rebuild_swap_chain(&self, window_id: window::Id) -> SwapChain {
```
mitchmindtree, line 756: I wonder if a better name for this might be `take_or_rebuild_swap_chain`? I think `get` tends to be used more often when retrieving a reference rather than taking ownership.

Also, perhaps this method should be private?

mitchmindtree (commented):
@danwilhelm thanks so much for getting to the bottom of this! Great find, very exciting to see this fixed :)

I left one comment which would be good to address, otherwise I'm more than happy to merge!

PR #725, opened 2021-04-15
Powerpreference default

mattheusroxas:
Changes `Powerpreference` to use `default`

### DIFF
--- a/nannou/src/wgpu/mod.rs
+++ b/nannou/src/wgpu/mod.rs
@@ -86,7 +86,7 @@ pub fn shader_from_spirv_bytes(
 }
 
 /// The default power preference used for requesting the WGPU adapter.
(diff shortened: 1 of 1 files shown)


mitchmindtree:
Context for future readers: this PR solves an issue where some systems will `panic!` during swapchain creation as a result of not being able to find a GPU that matches the `HighPerformance` power preference requested when instantiating the adapter.

By changing the default `PowerPreference` to `Default`, we increase the chance that we can find a compatible graphics device on the system.

In the future, we should:

1. Add a short tutorial to the nannou guide that shows how to specify the `HighPerformance` power preference and
2. Query for `HighPerformance`-capable devices in the case that wgpu ever gets an API that supports doing so.

mitchmindtree:
Related https://github.com/bevyengine/bevy/issues/402.

PR #726, opened 2021-04-15
Update wgpu to 0.7, conrod to 0.72, winit to 0.24.

### LINKED ISSUE #719: Crash on Macs when running the examples
**Workaround**
This issue has been reported several times on the Nannou Matrix chat. There is a temporary workaround:

> In `nannou/src/frame/mod.rs`, change the line `pub const DEFAULT_MSAA_SAMPLES: u32 = 4;` from 4 to 1.

---

**Problem**
Many examples (e.g. simple_window, wgpu_compute_shader) quickly crash on Macs, yielding the error `"No render pipeline state bound"`. Some example apps seem to work without a problem (e.g. wgpu_triangle), even with MSAA. These seem to (1) define their own render pipeline, and (2) do not call `frame.clear()`.

**Example**
`cargo run --example simple-window` crashes immediately with the error.

**The Workaround is not a Solution**
I am hesitant to provide the above workaround as a PR, because:
+ It would change the default behavior of existing apps; and 
+ 4x MSAA worked on versions of Nannou prior to Dec 9, 2020. It is also verified that Macs are capable of 4x MSAA via [the wpu-rs `msaa-line` example](https://github.com/gfx-rs/wgpu-rs/tree/master/examples/msaa-line) -- indicating a bug in Nannou.

**Findings**
Here are a few results of investigating this issue:
1. The bug seems to have been a side effect of moving from wgpu 0.5 to 0.6. A working commit is [here](https://github.com/nannou-org/nannou/tree/2cedce1014053c9771e592f8841d24006a05d9b2). Here is [the first non-working commit](https://github.com/nannou-org/nannou/tree/f2744d389b3b3708ee3071cf1fc2400b95a46f0d).
2. I believe the issue is caused by the render pass in [`resolve_texture`](https://github.com/nannou-org/nannou/blob/master/nannou/src/wgpu/mod.rs#L130). The examples work if this render pass is removed. 
3. I hypothesize that placing this render pass in a pipeline may resolve the issue -- perhaps move it into `Reshaper` as it originally was. (The Metal documentation seems to indicate vaguely that a render pipeline may be needed to perform MSAA.) 
4. The only two render functions that are not in pipelines are `clear_texture` and `resolve_texture`, both in `nannou/src/wpu/mod.rs` -- and both can cause crashes. (`clear_texture` causes a crash in examples that run `frame.clear()` -- even if they use a custom render pipeline.)

**Notes**
+ This [msaa-line PR](https://github.com/gfx-rs/wgpu-rs/pull/688) explains that MSAA values of 1 and 4 are the only universally accepted values. Further, there is currently no way in WebGPU to query the supported values.

### LINKED ISSUE #721: [Tracking Issue] Updating `wgpu` to `0.7`
Here is [the CHANGELOG entry](https://github.com/gfx-rs/wgpu-rs/blob/master/CHANGELOG.md#v07-2021-01-31) for wgpu-rs. Some exciting stuff in there like early WGSL support and better error messages which might help to solve issues like #719.

## Updating `conrod_wgpu`

The first step is to update `conrod_wgpu` to also target version 0.7. I've made a couple of short, failed attempts at this so far.

From memory, after running `cargo update`, one of `rusttype`'s dependencies had broken for some reason or another, so I tried updating `rusttype` itself. Unfortunately one of the APIs that `rusttype` provided was removed in the latest version as it was assumed no-one was using it :scream:

This lead me to refactor `conrod_core`'s text rendering a little to avoid needing the API, which turned out for the better and should result in slightly better performance. However, along with the removal of the API, another issue was introduced in `rusttype` that made it impossible to console the lifetime requirements between `PositionedGlyph`s yielded by the `Font::layout` and the GPU `Cache` type.

This has since been [fixed in master](https://gitlab.redox-os.org/redox-os/rusttype/-/merge_requests/158), however it turns out that `rusttype` has been [looking for a maintainer for the last year](https://gitlab.redox-os.org/redox-os/rusttype/-/issues/148) and it is unlikely these changes will be published until someone steps up. Unfortunately, I already have crates coming out my ears and don't think it would be healthy for me to take on another. *If anyone reading this would like to take it on please consider commenting at the repo!*

The previous `rusttype` maintainer alludes to a successor crate called `ab_glyph` which boasts slightly better performance. Both `conrod` and `nannou` lean on `rusttype` quite heavily for text layout, and I must admit that I'm not particularly excited about the idea of rewriting of conrod's text layout and rendering to use `ab_glyph` just to update from wgpu 0.6 to 0.7, particularly as I would like to eventually replace conrod with [`nannou_ui`](https://github.com/nannou-org/nannou/issues/383) and could be spending this time doing so.

I'm unsure of how to progress from here just yet. More than likely I'll end up raising my hand to help maintain `rusttype` temporarily... at least until we no longer depend on conrod for UI support.

mitchmindtree:
This consists of pretty simple name changes for the most part, however there is some pretty significant restructuring of the `RenderPipelineDescriptor`. This meant quite a few changes in the `render_pipeline_builder` module - I tried improving the builder method docs here while I was at it.

The `Sampler` binding type now requires that you specify whether or not it uses `Linear` filtering for any of its minify, magnify or mipmap filters. A `wgpu::sampler_filtering` function was added to make it a little easier to retrieve this bool from the `SamplerDescriptor` type. That said, this does require a little extra boilerplate for graphics pipelines that involve a texture sampler. I think the `wgpu_image.rs` example is probably the simplest example of how to update for this.

---

@danwilhelm @JoshuaBatty unfortunately (or maybe fortunately?) the new validation checks didn't turn up anything new. I wonder if maybe it was simply a bug in wgpu 0.6 that has since been fixed? Would you mind testing if #719 is still a problem on this branch?

---

@mattheusroxas the `PowerPreference::Default` variant was removed in the latest version of 0.7. I have a suspicion that the bug you were running into may have been fixed in this latest version of wgpu, so I've tried adding the `HighPerformance` preference back. Would you mind testing this branch? You can do so by changing your nannou dependency in your cargo.toml to this:
```toml
nannou = { git = "https://github.com/mitchmindtree/nannou", branch = "wgpu-0.7" }
```

---

Closes #721.
Closes #719.

### DIFF
--- a/nannou/src/draw/renderer/mod.rs
+++ b/nannou/src/draw/renderer/mod.rs
@@ -88,7 +88,7 @@ pub struct Renderer {
     text_bind_group_layout: wgpu::BindGroupLayout,
     text_bind_group: wgpu::BindGroup,
     texture_samplers: HashMap<SamplerId, wgpu::Sampler>,
-    texture_bind_group_layouts: HashMap<wgpu::TextureComponentType, wgpu::BindGroupLayout>,
+    texture_bind_group_layouts: HashMap<wgpu::TextureSampleType, wgpu::BindGroupLayout>,
     texture_bind_groups: HashMap<BindGroupId, wgpu::BindGroup>,
     output_color_format: wgpu::TextureFormat,
     sample_count: u32,
@@ -165,7 +165,7 @@ struct PipelineId {
     color_id: ColorId,
     alpha_id: AlphaId,
     topology: wgpu::PrimitiveTopology,
-    texture_component_type: wgpu::TextureComponentType,
+    texture_sample_type: wgpu::TextureSampleType,
 }
 
 impl Default for PrimitiveRender {
@@ -356,6 +356,8 @@ impl Renderer {
     pub const DEFAULT_GLYPH_CACHE_POSITION_TOLERANCE: f32 = 0.1;
     /// The texture format of the inner glyph cache.
     pub const GLYPH_CACHE_TEXTURE_FORMAT: wgpu::TextureFormat = wgpu::TextureFormat::R8Unorm;
+    /// The index format used to index into vertices.
+    pub const INDEX_FORMAT: wgpu::IndexFormat = wgpu::IndexFormat::Uint32;
 
     /// Create a new **Renderer**, ready to target an output attachment with the given size, sample
     /// count and color format.
@@ -390,7 +392,9 @@ impl Renderer {
         let fs_mod = wgpu::shader_from_spirv_bytes(device, include_bytes!("shaders/frag.spv"));
 
         // Create the glyph cache texture.
-        let text_sampler = wgpu::SamplerBuilder::new().build(device);
+        let text_sampler_desc = wgpu::SamplerBuilder::new().into_descriptor();
+        let text_sampler_filtering = wgpu::sampler_filtering(&text_sampler_desc);
+        let text_sampler = device.create_sampler(&text_sampler_desc);
         let glyph_cache_texture = wgpu::TextureBuilder::new()
             .size(glyph_cache_size)
             .usage(wgpu::TextureUsage::SAMPLED | wgpu::TextureUsage::COPY_DST)
@@ -427,7 +431,7 @@ impl Renderer {
             create_uniform_bind_group(device, &uniform_bind_group_layout, &uniform_buffer);
 
         // Bind group for text.
-        let text_bind_group_layout = create_text_bind_group_layout(device);
+        let text_bind_group_layout = create_text_bind_group_layout(device, text_sampler_filtering);
         let text_bind_group = create_text_bind_group(
             device,
             &text_bind_group_layout,
@@ -592,27 +596,27 @@ impl Renderer {
                     }
 
                     // Retrieve the current texture view and texture view ID. These are necessary
-                    // for producing the curren tpipeline and bind group IDs. Also ensure we have
+                    // for producing the current pipeline and bind group IDs. Also ensure we have
                     // an entry for them in our map.
                     let tex_view = match render.texture_view {
                         Some(tex_view) => tex_view,
                         None => self.default_texture_view.clone(),
                     };
                     let tex_view_id = tex_view.id();
-                    let texture_component_type = tex_view.component_type();
+                    let texture_sample_type = tex_view.sample_type();
                     new_tex_views.insert(tex_view_id, tex_view);
 
                     // Determine the new current bind group layout ID, pipeline ID, bind group ID
                     // and scissor required for drawing this primitive.
                     let new_pipeline_id = {
-                        let color_id = blend_descriptor_hash(&curr_ctxt.color_blend);
-                        let alpha_id = blend_descriptor_hash(&curr_ctxt.alpha_blend);
+                        let color_id = blend_state_hash(&curr_ctxt.color_blend);
+                        let alpha_id = blend_state_hash(&curr_ctxt.alpha_blend);
                         let topology = curr_ctxt.topology;
                         PipelineId {
                             color_id,
                             alpha_id,
                             topology,
-                            texture_component_type,
+                            texture_sample_type,
                         }
                     };
                     let new_bind_group_id = {
@@ -643,7 +647,11 @@ impl Renderer {
                         curr_pipeline_id = Some(new_pipeline_id);
                         let color_blend = curr_ctxt.color_blend.clone();
                         let alpha_blend = curr_ctxt.alpha_blend.clone();
-                        new_pipeline_ids.insert(new_pipeline_id, (color_blend, alpha_blend));
+                        let sampler_filtering = wgpu::sampler_filtering(&curr_ctxt.sampler);
+                        new_pipeline_ids.insert(
+                            new_pipeline_id,
+                            (color_blend, alpha_blend, sampler_filtering),
+                        );
                         let cmd = RenderCommand::SetPipeline(new_pipeline_id);
                         self.render_commands.push(cmd);
                     }
@@ -700,12 +708,16 @@ impl Renderer {
         // Clear new combos that we already have.
         new_pipeline_ids.retain(|id, _| !self.pipelines.contains_key(id));
         // Create new render pipelines as necessary.
-        for (new_id, (color_blend, alpha_blend)) in new_pipeline_ids {
+        for (new_id, (color_blend, alpha_blend, sampler_filtering)) in new_pipeline_ids {
             let bind_group_layout = self
                 .texture_bind_group_layouts
-                .entry(new_id.texture_component_type)
+                .entry(new_id.texture_sample_type)
                 .or_insert_with(|| {
-                    create_texture_bind_group_layout(device, new_id.texture_component_type)
+                    create_texture_bind_group_layout(
+                        device,
+                        sampler_filtering,
+                        new_id.texture_sample_type,
+                    )
                 });
             let new_pipeline = create_render_pipeline(
                 device,
@@ -744,7 +756,7 @@ impl Renderer {
             let texture_view = &new_tex_views[&new_tex_view_id];
             // Retrieve the associated bind group layout.
             let bind_group_layout =
-                &self.texture_bind_group_layouts[&pipeline_id.texture_component_type];
+                &self.texture_bind_group_layouts[&pipeline_id.texture_sample_type];
             // Create the bind group.
             let bind_group =
                 create_texture_bind_group(device, bind_group_layout, sampler, texture_view);
@@ -884,7 +896,7 @@ impl Renderer {
         let mut render_pass = render_pass_builder.begin(encoder);
 
         // Set the buffers.
-        render_pass.set_index_buffer(index_buffer.slice(..));
+        render_pass.set_index_buffer(index_buffer.slice(..), Self::INDEX_FORMAT);
         render_pass.set_vertex_buffer(0, point_buffer.slice(..));
         render_pass.set_vertex_buffer(1, color_buffer.slice(..));
         render_pass.set_vertex_buffer(2, tex_coords_buffer.slice(..));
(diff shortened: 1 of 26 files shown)


mattheusroxas:
Yeah doesn't work. Returns the same driver error as before:
```
thread 'main' panicked at 'internal error: entered unreachable code: Unexpected result - driver bug? Err(ERROR_UNKNOWN)', /home/mattheus/.local/share/cargo/registry/src/github.com-1ecc6299db9ec823/gfx-backend-vulkan-0.7.0/src/device.rs:1953:18
stack backtrace:
   0: rust_begin_unwind
             at /rustc/2fd73fabe469357a12c2c974c140f67e7cdd76d0/library/std/src/panicking.rs:493:5
   1: core::panicking::panic_fmt
             at /rustc/2fd73fabe469357a12c2c974c140f67e7cdd76d0/library/core/src/panicking.rs:92:14
   2: gfx_backend_vulkan::device::<impl gfx_backend_vulkan::Device>::create_swapchain
   3: <gfx_backend_vulkan::window::Surface as gfx_hal::window::PresentationSurface<gfx_backend_vulkan::Backend>>::configure_swapchain
   4: wgpu_core::device::<impl wgpu_core::hub::Global<G>>::device_create_swap_chain
   5: <wgpu::backend::direct::Context as wgpu::Context>::device_create_swap_chain
   6: wgpu::Device::create_swap_chain
   7: nannou::window::Builder::build
   8: nannou::app::SketchBuilder<E>::run
   9: art::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

danwilhelm:
It seems there is now only `LowPower` and `HighPerformance` (or selecting an adapter manually from the possibilities). Perhaps a temporary solution might be to write an informative `panic!` if the swap chain creation fails, asking the user to try the other option.

Here is the relevant `wgpu` PR: https://github.com/gfx-rs/wgpu/commit/c933b9766a404c54f0773b006138d8fb7d3a637f

yutannihilation:
The change seems to be because `Default` was just an alias to `LowPower`. So, isn't it safe to use `LowPower` instead of `HighPerformance`?

> It's confusing and not very useful if it's just an alias to `LowPower`.
https://github.com/gfx-rs/wgpu/issues/971

nannou/src/draw/renderer/mod.rs
```diff
@@ -356,6 +356,8 @@ impl Renderer {
     pub const DEFAULT_GLYPH_CACHE_POSITION_TOLERANCE: f32 = 0.1;
     /// The texture format of the inner glyph cache.
     pub const GLYPH_CACHE_TEXTURE_FORMAT: wgpu::TextureFormat = wgpu::TextureFormat::R8Unorm;
+    /// The index format used to index into vertices.
+    pub const INDEX_FORMAT: wgpu::IndexFormat = wgpu::IndexFormat::Uint32;
```
danwilhelm, line 360: Good catch!

nannou/src/wgpu/render_pipeline_builder.rs
```diff
@@ -428,43 +508,45 @@ fn build(
...
+    //         Some(rasterization_state.unwrap_or(RenderPipelineBuilder::DEFAULT_RASTERIZATION_STATE))
+    //     }
+    //     false => {
+    //         if rasterization_state.is_some() {
+    //             panic!("specified rasterization state fields but no fragment shader");
+    //         } else {
+    //             None
+    //         }
+    //     }
+    // };
```
danwilhelm, line 540: Were these comments intended to be left in?

nannou/src/wgpu/render_pipeline_builder.rs
```diff
@@ -28,56 +28,46 @@ pub struct RenderPipelineBuilder<'a> {
...
-    vertex_buffers: Vec<wgpu::VertexBufferDescriptor<'static>>,
-    sample_count: u32,
-    sample_mask: u32,
-    alpha_to_coverage_enabled: bool,
+    primitive: wgpu::PrimitiveState,
+    color_state: Option<wgpu::ColorTargetState>,
+    color_states: &'a [wgpu::ColorTargetState],
+    depth_stencil: Option<wgpu::DepthStencilState>,
+    vertex_buffers: Vec<wgpu::VertexBufferLayout<'static>>,
+    multisample: wgpu::MultisampleState,
```
danwilhelm, line 36: Nice job figuring out how to update the `RenderPipeline` (not just these lines but throughout)!

nannou/src/wgpu/sampler_builder.rs
```diff
@@ -20,6 +20,7 @@ impl<'b> SamplerBuilder {
     pub const DEFAULT_COMPARE: Option<wgpu::CompareFunction> = None;
     pub const DEFAULT_ANISOTROPY_CLAMP: Option<NonZeroU8> = None;
     pub const DEFAULT_LABEL: Option<&'static str> = None;
```
danwilhelm, line 22: A default label is defined for `TextureView`, but not here -- should these be made consistent?

`pub const DEFAULT_LABEL: &'static str = "nannou-texture-view";`

nannou/src/wgpu/texture/mod.rs
```diff
@@ -751,23 +751,7 @@ pub fn data_size_bytes(desc: &wgpu::TextureDescriptor) -> usize {
...
-        | Rgb10a2Unorm | Rg11b10Float => 4,
-
-        Rg32Uint | Rg32Sint | Rg32Float | Rgba16Uint | Rgba16Sint | Rgba16Float | Rgba32Uint
-        | Rgba32Sint | Rgba32Float => 8,
-
-        Depth32Float | Depth24Plus | Depth24PlusStencil8 => 4,
-
-        _ => unimplemented!(),
-    }
+    format.describe().block_size as u32
```
danwilhelm, line 754: Nice find! The docs mention this might only return the same thing on uncompressed textures (not sure if this is relevant for `nannou`) -- https://docs.rs/wgpu-types/0.7.0/wgpu_types/struct.TextureFormatInfo.html

danwilhelm (approved):
Nice job!

mitchmindtree:
@JoshuaBatty would you mind testing this again on macos when you get a chance but with `.power_preference(wgpu::PowerPreference::LowPower)` on the window constructor? I'm curious if you still get that epic system crash like you did on 0.6.

> Perhaps a temporary solution might be to write an informative panic! if the swap chain creation fails, asking the user to try the other option.

@danwilhelm good idea, and thanks for the review! I'll address these tomorrow.

> change seems to be because Default was just an alias to LowPower. So, isn't it safe to use LowPower instead of HighPerformance?

@yutannihilation you're right to point out `Default` being equivalent to `LowPower`. For some context, yesterday we landed a PR that changed the default power preference from `HighPerformance` to `Default` as it fixed a `panic!` that was occurring for one user #725 and we figured it to be a safer default for nannou generally. However, later that day @JoshuaBatty pulled down the master branch and it turned out that switching to `Default` caused macOS to crash quite spectacularly.

Both this, and the name change of the variant from `Default` to `LowPower` indicate to me that the `LowPower` doesn't necessarily seem to be a safer default, and that we might as well opt for `HighPerformance` as it seems a better suited option for general applications of nannou (fancy sketches, graphics-intensive stuff). I think ultimately wgpu's `PowerPreference` API is intended to be just that - a *preference*, and in the case that either variant is not available on the system, wgpu should in theory fall back to what is available anyway (which is clearly what it's trying to do [here](https://github.com/frbimo/wgpu/blob/fdd561e69e3256409e3ae8fe5e0b81c64b6d188e/wgpu-core/src/instance.rs#L504-L505). The reality however is that whatever option you choose, it may successfully result in choosing a different GPU, however there's no guarantees that the selected driver isn't far buggier than the other one, or that the low level gfx backend implementation is not yet handling a certain case for that driver, or that the driver doesn't support some feature that wgpu requires and it doesn't trigger a crash until you're creating the swapchain / texture / render pipeline, etc. 

I think @danwilhelm's suggestion makes sense, where we choose the preference that makes sense for nannou, but try to indicate to users that they can try changing it in the case that they do run into issues. I think the tricky part will be reliably detecting when an error is caused by a "bad" power preference... e.g. @mattheusroxas' error didn't occur until swapchain creation, which is not necessarily related to selecting a GPU based on power preference at all.

One small step we can do is document `PowerPreference` in the guide and the associated window builder method with a short tutorial. In that case we at least have somewhere to point to that explains everything in the case a user does have some related issue.

danwilhelm:
@mindtree FWIW changing the default to `LowPower` worked on my (integrated GPU) Mac, both the tests and selected examples. The tests/examples also worked on my PC.

yutannihilation:
@mitchmindtree 
Thanks for the details, I only saw #725 and didn't follow the other discussions. I almost agree with you. One minor thing, I feel this part

> and the name change of the variant from `Default` to `LowPower` indicate to me that the `LowPower` doesn't necessarily seem to be a safer default

might not be true as `LowPower` stays as `Default` trait.

https://github.com/gfx-rs/wgpu/blob/6a6a9a4aedbbf7c8cc6b11a4b0c42341ac50a798/wgpu-types/src/lib.rs#L88-L92

JoshuaBatty:
Hey @mitchmindtree i just tried running some examples with `.power_preference(wgpu::PowerPreference::LowPower)` on the window constructor and I'm not getting that system crash on my mac anymore :)

nannou/src/wgpu/texture/mod.rs
mitchmindtree: Ahhh good point! Hmmm it might be worth opening a dedicated issue and doing a review of the use of `format_size_bytes` throughout nannou and how compressed textures might affect handling certain cases in nannou's "extended" texture API. I haven't personally used compressed textures yet, but have been curious about them for a while, especially for use in real-time video decoding formats that support random access like HAP.

mitchmindtree:
@danwilhelm @JoshuaBatty thanks for testing, glad to hear there's no more epic crash!

> might not be true as LowPower stays as Default trait.

Good point - I asked on the wgpu matrix if there was a reason for this but there doesn't seem to be other than that having a default is useful. Will report back if I find out anything that contradicts this.

After chatting with kvark on matrix, it's likely that the panic @mattheusroxas is running into is a bug in `gfx-backend-vulkan`. I've opened https://github.com/gfx-rs/gfx/issues/3727 to track this. In the meantime, I think I'm going to keep the default power preference `HighPerformance` for the reasons mentioned above, at least until we find out a bit more about what's going on in that issue.

PR #727, opened 2021-04-19
Add ability to color characters individually, update draw_text example

danwilhelm:
Adds a new method that colors text glyphs individually. 

For example, this could be useful for:
- Quickly colorizing the first character or word of a string. 
- Displaying syntax-highlighted code.

For an example, see `draw_text.rs`. Tested by running the examples, `cargo test`, and `cargo fmt`.


mitchmindtree:
Very cool! Looks good to me, happy to land this.

PR #732, opened 2021-05-03
nannou: more helpful error messages and docs.

digitalatigid:
I ran across this error and just needed someone to tell me to check my
drivers, so now nannou can do this.

### DIFF
--- a/nannou/src/window.rs
+++ b/nannou/src/window.rs
@@ -156,10 +156,21 @@ pub type UnfocusedFn<Model> = fn(&App, &mut Model);
 pub type ClosedFn<Model> = fn(&App, &mut Model);
 
 /// Errors that might occur while building the window.
-#[derive(Debug)]
 pub enum BuildError {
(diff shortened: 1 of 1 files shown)


mitchmindtree:
Thanks for this @digitalatigid !

After talking to the wgpu folks, I believe in theory the `PowerPreference` *should* actually behave like a preference. E.g. if you only have an integrated GPU but you select `HighPerformance`, wgpu will fall back to `LowPower` without issues. I believe most issues that are arising are on multi-GPU systems where wgpu may indeed select a different device based on the specified preference, however it might turn out wgpu itself does not yet fully support the card, or doesn't quite have the ability to detect if the card has the necessary capabilities to be used before selecting it. I hope that makes some kind of sense.

All that said, the `NoAvailableAdapter` normally occurs if *no* device can be found at all - this is more often due to missing the necessary drive (often the Vulkan driver on Linux for example). Can you confirm that changing power preference was what solved your `NoAvailableAdapter` error, and that it wasn't solved by possibly installing a missing graphics driver?

Issue #691, opened 2020-12-18
Example from guide build fails


bcbroom:
I'm unable to run the example `cargo run --release --example draw` on macOS 10.15.7 (Catalina), and get

```   
Compiling glsl-to-spirv v0.1.7
error: failed to run custom build command for `glsl-to-spirv v0.1.7`

Caused by:
  process didn't exit successfully: `/Users/broom/Documents/code/rust/nannou/target/release/build/glsl-to-spirv-6de94febe694eea4/build-script-build` (exit code: 101)
  --- stdout
  cargo:rerun-if-changed=build/glslangValidator.exe
  running: "cmake" "/Users/broom/.cargo/registry/src/github.com-1ecc6299db9ec823/glsl-to-spirv-0.1.7/glslang" "-DCMAKE_INSTALL_PREFIX=/Users/broom/Documents/code/rust/nannou/target/release/build/glsl-to-spirv-f1f9973f393be420/out" "-DCMAKE_C_FLAGS= -ffunction-sections -fdata-sections -fPIC -m64 -arch x86_64" "-DCMAKE_C_COMPILER=/usr/bin/cc" "-DCMAKE_CXX_FLAGS= -ffunction-sections -fdata-sections -fPIC -m64 -arch x86_64" "-DCMAKE_CXX_COMPILER=/usr/bin/c++" "-DCMAKE_ASM_FLAGS= -ffunction-sections -fdata-sections -fPIC -m64 -arch x86_64" "-DCMAKE_ASM_COMPILER=/usr/bin/cc" "-DCMAKE_BUILD_TYPE=Release"

  --- stderr
  fatal: not a git repository (or any of the parent directories): .git
  thread 'main' panicked at '
  failed to execute command: No such file or directory (os error 2)
  is `cmake` not installed?

  build script failed, must exit now', /Users/broom/.cargo/registry/src/github.com-1ecc6299db9ec823/cmake-0.1.45/src/lib.rs:894:5
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
warning: build failed, waiting for other jobs to finish...
error: build failed
```

I also tried
```
broom@Brian-Secondary-MBP nannou % rustup update
info: syncing channel updates for 'stable-x86_64-apple-darwin'
info: checking for self-updates

  stable-x86_64-apple-darwin unchanged - rustc 1.48.0 (7eac88abb 2020-11-16)

info: cleaning up downloads & tmp directories
[...]

PR #741, opened 2021-05-19
Issue #691 Adding cmake to OS X setup directions

satchm0h:
Per issue #691, the initial example command in the getting started guide fails to build if `cmake` has not been installed. I've just added simple directions to install cmake using [Homebrew](https://brew.sh)


mitchmindtree:
Thanks!

PR #745, opened 2021-06-06
Fix `glyph_colors` miscoloring on resize

### LINKED ISSUE #744: `glyph_colors` miscolors on resize
In the `draw_text` example, when the window is resized causing `Lorem` and `ipsum` to line break, the i in ipsum becomes black instead of red.

This is caused because `positioned_glyphs` skips white space text when it adds line breaks.

Hence, this causes the vector of glyph colors to be misaligned.

```
'GGGGGKRRRRR'
'Lorem ipsum'
=>
'GGGGGKRRRRR'    # Same colors
'Loremipsum'     # Space is no longer rendered due to line break
      ^ i is now black
```

danwilhelm:
Fix #744

Uses `text.line_infos()` to skip glyph colors corresponding to non-rendered characters.

More code than I hoped, but takes advantage of `Info`s being ordered by ascending character position.

I did try a two-liner solution (~included in the comments), but it isn't efficient when many lines are rendered. Any suggestions for a more concise solution would be appreciated!

### DIFF
--- a/nannou/src/draw/primitive/text.rs
+++ b/nannou/src/draw/primitive/text.rs
@@ -366,12 +366,19 @@ impl draw::renderer::RenderPrimitive for Text<f32> {
             geom::Rect::from_corners(geom::pt2(l, b), geom::pt2(r, t))
         };
 
-        // Extend the mesh with a rect for each displayed glyph.
-        // If `glyph_colors` contains insufficient colors, use `color` for the remaining glyphs.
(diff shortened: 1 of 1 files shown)


danwilhelm:
I found a much more concise/elegant solution and re-pushed! For efficiency, this still assumes that the text `Info`s are in ascending character order.

mitchmindtree:
Hey looks good to me, nice job!

PR #749, opened 2021-06-10
fix(docs): correct the triangle_mode docs

QQism:
Hi, I've just noticed the documentation for `Draw#triangle_mode` is not correct. This commit will fix it.


mitchmindtree:
Good spot, thanks!

PR #750, opened 2021-06-13
Ensure `Info` char index refers to orig string

danwilhelm:
When wrapping text, `Info` (and the enclosed `Break`) records the byte and character indexes of each wrapped line.

**Problem.** Currently, the byte index refers to the original string, but the char index does not. For example:

```
let s1 = "fn main() {\n    nannou::app(model)\n";              // Byte index is into original string
let s2 = "fn\n main(\n) {\n    \nnanno\nu::app\n(mode\nl)\n";  // Char index is into "wrapped" string
```

**Proposed Solution.** I propose making the byte and character indexes both refer to the original string, to avoid confusion. Currently, the char index is always incremented by 1 every break, even if the break itself is 0 chars (`""`), 1 char (`" "`, `"\n"`), or 2 chars (`"\r\n"`). It does not skip these end-of-break chars, even though the byte index skips them.

**Consequence.** Changing this will help the text coloring implementation, which assumes that text colors are ordered by character index into the original string. It is very difficult to tell the character->line mappings as it is currently implemented.

---

**Example.**

Suppose we have a string:

`let s = "fn main() {\n    nannou::app(model)\n";`

We then wrap this across numerous lines:

```
"fn"      <- Info { start_byte: 0, start_char: 0, end_break: Wrap { byte: 2, char: 2, len_bytes: 1 }, ...
"main("   <- Info { start_byte: 3, start_char: 3, end_break: Wrap { byte: 8, char: 8, len_bytes: 0 }, ...
") {"     <- Info { start_byte: 8, start_char: 9, end_break: Newline { byte: 11, char: 12, len_bytes: 1 }, ...
"   "     <- Info { start_byte: 12, start_char: 13, end_break: Wrap { byte: 15, char: 16, len_bytes: 1 }, ...
"nanno"   <- Info { start_byte: 16, start_char: 17, end_break: Wrap { byte: 21, char: 22, len_bytes: 0 }, ...
"u::app"  <- Info { start_byte: 21, start_char: 23, end_break: Wrap { byte: 27, char: 29, len_bytes: 0 }, ...
"(mode"   <- Info { start_byte: 27, start_char: 30, end_break: Wrap { byte: 27, char: 29, len_bytes: 0 }, ...
"l)"      <- Info { start_byte: 32, start_char: 36, end_break: Newline { byte: 34, char: 38, len_bytes: 1 }, ...
```

We observe the following from the above, which is not typical:
- `start_char > start_byte`.
- `start_char > s.chars.count()`.

Effectively, this `start_char` represents the character position in the following string, which is longer than the original: 
```
"fn\n main(\n) {\n    \nnanno\nu::app\n(mode\nl)\n"
```

To be consistent with `start_byte`, `start_char` should give the index into the characters of the original string.

### DIFF
--- a/nannou/src/draw/primitive/text.rs
+++ b/nannou/src/draw/primitive/text.rs
@@ -248,8 +248,10 @@ where
         self.map_ty(|ty| ty.layout(layout))
     }
 
-    /// Set a color for each glyph.
+    /// Set a color for each glyph, which is typically one character.
     /// Colors unspecified glyphs using the drawing color.
+    /// NOTE: Sometimes, a glyph can represent multiple characters,
+    ///       or be a part in other glyphs.
     pub fn glyph_colors<I, C>(self, glyph_colors: I) -> Self
     where
         I: IntoIterator<Item = C>,
--- a/nannou/src/text/line.rs
+++ b/nannou/src/text/line.rs
@@ -17,6 +17,9 @@ pub enum Break {
         /// The byte length which should be skipped in order to reach the first non-whitespace
         /// character to use as the beginning of the next line.
         len_bytes: usize,
+        /// The number of chars which should be skipped in order to reach the first non-whitespace
+        /// character to use as the beginning of the next line.
+        len_chars: usize,
     },
     /// A break caused by a newline character.
     Newline {
@@ -26,6 +29,8 @@ pub enum Break {
         char: usize,
         /// The width of the "newline" token in bytes.
         len_bytes: usize,
+        /// The width of the "newline" token in chars.
+        len_chars: usize,
     },
     /// The end of the string has been reached, with the given length.
     End {
@@ -56,7 +61,7 @@ pub struct NextBreak {
 pub struct Info {
     /// The index into the `&str` that represents the first character within the line.
     pub start_byte: usize,
-    /// The character index of the first character in the line.
+    /// The character index into the `&str` of the first character in the line.
     pub start_char: usize,
     /// The index within the `&str` at which this line breaks into a new line, along with the
     /// index at which the following line begins. The variant describes whether the break is
@@ -238,7 +243,8 @@ fn next_break(text: &str, font: &text::Font, font_size: FontSize) -> NextBreak {
                 let break_ = Break::Newline {
                     byte: byte_i,
                     char: char_i,
-                    len_bytes: 2,
+                    len_bytes: 2, // skip "\r\n"
+                    len_chars: 2,
                 };
                 return NextBreak {
                     break_,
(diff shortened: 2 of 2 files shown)


mitchmindtree:
Thanks for digging in @danwilhelm, I haven't had a chance to test this myself but I think I see where you're coming from.

You're right that the original intention for the `start_char` index was that calling `s.chars().nth(start_char)` should give you the character beginning after the line break. I also think you're right to look at the `info.start_char + char + 1` as the source of the issue. Specifically, it looks like the `+ 1` only makes sense if the original string contained a line break character (i.e. `\n`, `\r\n`) but *not* if the line break is caused by a `Wrap`.

I'm a little unsure how replacing the `1` with `len_bytes` fixes the issue here? Particularly as `char` index and byte counts are very different kinds of indices (keeping in mind that `char` index refers to the index of the character within a UTF8 string some of which span one byte and some of which span multiple bytes, whereas byte indices are the raw byte indices into the original string slice).

danwilhelm:
The issue I'm having is that the text colors passed in `glyph_colors` end up mapping to the characters of the original string.

To get the correct colors for each wrap, I need to map each `Info` span back to the character indexes of the original string. I can do this accurately if `start_char` refers to the original string. Otherwise, I'd have to duplicate the wrapping algorithm to get the correct char positions.

I am adding `len_bytes` here because given the two current wrapping methods, each byte skipped is exactly one character -- "" is zero characters, "\n" or " " is one character, and "\r\n" is two characters. So to accurately increment the character position in the original string, you can add `len_bytes`. (The same as is done when incrementing the byte position.)

mitchmindtree:
> I am adding len_bytes here because given the two current wrapping methods, each byte skipped is exactly one character -- "" is zero characters, "\n" or " " is one character, and "\r\n" is two characters. So to accurately increment the character position in the original string, you can add len_bytes -- given the two current wrapping methods.

Ahh yes I see the issue!

I think we should still be careful not to use `len_bytes` here though, as it's only really a coincidence that the `len_bytes` to skip is equal to the number of characters to skip.

I think a "proper" fix might be to add a `len_chars` field to the `Wrap` and `Newline` variants which describes the number of `char`s that must be skipped.

Alternatively, during iteration we could use the indices that we already have to index into the part of the string that we want to skip and call `s.chars().count()` to determine the length in `char`s that way.

While `len_bytes` happens to work in this case (as each whitespace or newline `char` we track happens to only be one byte when encoded in utf8), I think the mismatch between lengths in bytes and chars can be a little confusing to read and would break if in the future we add a check for some other multi-byte newline utf8 character for example (although that's probably unlikely?).

danwilhelm:
I agree, adding a new `len_chars` field seems like the best approach. I just updated the commit, along with some extra comments explaining that 1 glyph is typically 1 char.

Note that the break detection functions loop through character-by-character, so we are assured that 1 byte == 1 char in these particular two functions.

PR #751, opened 2021-06-14
fix: NOC 3_04 exercise to work on Windows

QQism:
When running the 3_04 exercise on Windows, the background is completely
black, so the spiral path is not visible. I tested on my mac, the
background, without being set, is green, so it kinda works on macOS.
Nevertheless, this commit sets the background to white on the first
frame so it can work on both Windows and macOS and be consistent with
the Nature of Code's example, which has the white background as well.

### Before (On Windows)
nannou - 3_04_exercise_spiral 14_06_2021 12_33_59 PM

### After (On Windows)
nannou - 3_04_exercise_spiral 14_06_2021 12_02_39 PM

### Before (on macOS)
Screen Shot 2021-06-14 at 12 18 35 PM

### After (On macOS)
Screen Shot 2021-06-14 at 12 38 49 PM


mitchmindtree:
Thanks for the patch!

PR #753, opened 2021-06-15
Enable serializing of color types

BenBergman:
Enabling palette's `serializing` feature allows colors to be saved to config files easily.


mitchmindtree:
Ah nice catch, thanks!

Issue #387, opened 2019-08-18
Improving iterative compile times for projects depending on nannou


mitchmindtree:
Slightly related to #178 but more focused on iterative compile times than fresh ones.

Currently it's easy to reach fairly significant compile times even for small projects (e.g. 10+ secs). I think there's likely a lot we could do to improve this situation with the design of nannou itself.

Here are some steps I think we could take:

- [ ] First, setup a reliable iterative compilation time bench-marking process. This should have some way of measuring iterative re-compile times for many different size projects including both examples and more significant downstream projects like [spatial_audio_server](https://github.com/museumsvictoria/spatial_audio_server). Just getting a setup working for examples would be a good start. I believe rustc provides a way to get a breakdown of exactly where the compiler is spending most of its time, though I don't remember how to do this off the top of my head.
- [ ] Survey the surface-level of nannou's API. Attempt to remove generic types in favour of dynamic dispatch where possible. This should be more than doable in many of the top-level APIs without any significant performance penalty.
- [ ] Encourage code generation during library compile-time where possible. E.g. perhaps providing type and function wrappers around generic types and functions respectively will assist the compiler to pre-generate a lot of code for the common use case?

Any other ideas are welcome!

I've labelled this as good-first-issue as this might be a nice way to become familiar with a lot of nannou's internals and rust benchmarking tools.

mitchmindtree:
Some interesting discussion on this issue popping up [here](https://www.reddit.com/r/rust/comments/ctlt16/thoughts_on_rust_bloat/).

michal-z:
On Windows using LLVM lld-link.exe instead of MS linker improves build times. Please see here: https://github.com/rust-gamedev/wg/issues/50

mystal:
On another project I'm working on, the biggest compile time improvements were:
1. Use `lld` to link on supported platforms.
2. For debug builds, set `debug = false` to remove all debug info from the binaries. I rarely use a debugger to debug Rust, and having this turned on increases the build time quite a bit! So I have it off by default and toggle it on if I ever need to attach a debugger.
3. Deduplicate dependencies when possible--i.e. only have one version of each dependency, especially `syn`/`quote`. This is less of a problem for incremental compilation, but I think it helps.

kazimuth:
See also the "enable fast compiles" section of the Bevy setup guide: https://bevyengine.org/learn/book/getting-started/setup/

Issue #209, opened 2018-11-04
Create a "headless" example (an example that demonstrates an application with no windows)


mitchmindtree:
We shouldn't make the assumption that the user always wants one or more windows - there are plenty of applications where the user might only want to use audio, lasers, etc.

This will involve tweaking some internals, and perhaps separating all of the windowing logic and user input event loop from the `App` type. I've been meaning to do a refactor of the `App` type so it could be useful to consider this issue while doing so.

mitchmindtree:
It would be nice if nannou could still provide raw keyboard and mouse events in a "headless" context, as these might be useful for interacting with alternative displays like LED arrays, laser projectors, etc. This would require creating a `winit::EventLoop` instance, but without actually building any windows. I know that some backends support raw device events (device events unassociated with particular windows) but I haven't tested each of them to know for sure. Perhaps the `EventLoop` can be lazily initialized only if one of these input functions are requested? In the case that no input function is requested, we can assume the `EventLoop` is not needed.

I imagine the API would still be quite similar - the `model` function would still be necessary for setup and `Model` initialisation, and would allow for spawning these alternate outputs etc.

The vulkan instance should still be available in case the user wishes to do some form of image processing or GPU compute.

Songtronix:
How hard would it be to implement a very basic headless example?

tmladek:
Hello, any news on this issue? :)

From my reading of the docs and a couple brief attempts, it would seem that nannou doesn't yet support headless `App`s...

mitchmindtree:
Hey @tmladek, that's correct, we're still in the same state. My previous comment mentions some ideas on how we might be able to make some progress here if anyone is interested in contributing. Otherwise I'm sure we'll get around to this at some point, it just hasn't had high enough priority for us to focus on this just yet.

Another idea - Before trying to work on a solution in nannou itself, I think I would first try to land a `headless.rs` example in `winit` (the library nannou uses to manage windowing and input device events) that demonstrates a program that continues to receive device input (e.g. mouse and keyboard), but does not need a window to do so. Once that can be shown to work nicely in a cross platform manner, it should be trivial to land headless support in nannou.

merlindru:
@bitzl Shouldn't that be possible already? You'd advance by one frame for each of your video, capture it (see #187), and then stitch it together into a video file

melMass:
@merlindru This still requires a window or I missed a point in the link?

Issue #178, opened 2018-07-16
Reduce build times for both `nannou` and projects created via `nannou-new`


mitchmindtree:
After running a few workshops, we've run into a couple of serious bottlenecks for new users getting started with nannou.

*Note: This issue does not refer to automation of the "getting started" process, just the build times themselves. I'll open a separate issue for this.*

## 1. Initial build time of `nannou`

After installing the rust toolchain, the first step is to clone the nannou repo and build it in order to run the examples. For users with relatively new machines, this takes around 3-5 minutes from scratch, which is already a long time in itself, however for users with older machines it can take up to 20mins (maybe longer)? This is a serious time-sink for workshops where we just want to get users running the examples and playing as quickly as possible.

**Possible Solution**

Ideally, we'd distribute pre-built versions of the library that are hashed against:

- the target triplet (architecture, OS)
- the latest, stable version of rustc and cargo used to build them.
- the version of nannou itself

Ideally, if a build does not exist with this hash, or if some sort of unexpected error occurs when attempting to link this crate, the build process would then fall back to the default cargo procedure of building from source.

An even more awesome solution would be if cargo itself decided to start shipping pre-built binaries, however this would obviously massively increase the cost due to requiring time and access to all platforms necessary for pre-building and hashing the binaries (whereas currently they just need to store and serve source code).

Having a system for building nannou for multiple platforms, storing and serving these binaries is also something that we will also have to solve, though is obviously much more achievable than doing so for all 10k+ crates or whatever the number is now currently on cargo.

Related issues:

- rust-lang/cargo#1139

## 2. Build times of new sketches/apps generated via `nannou-new`

Cargo currently requires re-building the entire `nannou` package for each newly created "sketch"/"app" generated via the `nannou-new` tool. This is obviously a massive time-sink for all the reasons described above, though is particularly troublesome for a couple reasons:

- The idea of `nannou-new` is to allow the user to get started with sketching an idea as quickly and efficiently as possible. Waiting 20 minutes is easily enough time to forget about your original idea plus 20 others in the time it just takes to compile nannou again.
- The frequency of building new sketches is far more common than building nannou itself. Users are much more likely to just open a new processing file if just creating the nannou project itself take 20+ mins.

**Possible Solution**

[...]

PR #754, opened 2021-06-18
Add a `nannou_core` crate. Switch from `cgmath` to `glam`. Remove `geom::Graph`. Remove generic Scalar param from `Draw` API.

### LINKED ISSUE #752: Abstract a `nannou_core` crate for use in `no_std` or headless environments
@JoshuaBatty and I have often come across the desire to use nannou in headless or no_std environments, but without the deep stack of crates required for providing the event loop, wgpu context, etc. These include:

- Using nannou to generate patterns for LASERs.
- Using nannou in hot-loaded dynamic libraries.
- Sharing nannou types with embedded devices that are a part of an art installation.
- Using nannou types in `rust-gpu` code.

I'm thinking `nannou_core` could be home to those abstractions that don't directly interact with I/O, e.g.

- The `geom` module (replacing our existing `Point`/`Vector` types with `glam`).
- The `noise` module if possible... In theory it should be possible to include this, but we will have to investigate the feasibility of support for `no_std`. https://github.com/Razaekel/noise-rs/issues/251.
- The `rand` module with simple helper functions extending the `rand` crate.
- The `color` module with `palette` re-exported along with our helpers.
- The `ease` module (`pennereq` requires a `#![no_std]` feature, but should be a trivial contribution).
- The `text` module with logic for laying out text. Would require `#![no_std]` support in `rusttype` which is currently unmaintained... we might have to address this anyway by switching to an alternative.

We could also add some more granular feature gates for these (e.g. `serde`, `rand`) along with a `nannou_core`-specific prelude.

Then, I imagine `nannou` would re-export the `nannou_core` modules resulting in an interface that is largely indistinguishable from the current one.

### `rust-gpu` support

We might be able to also add a dedicated CI job that checks `nannou_core` against the `spirv-unknown-unknown` target to ensure we maintain compatibility with `rust-gpu`.

### `nannou_draw`?

Possibly this might lead to a future where we can also separate out a `nannou_draw` crate that allows for using `Draw` to render directly to images or vectors in memory (i.e. svg, png, laser graphics, etc), without needing the whole winit+wgpu stack.

mitchmindtree:
This is a large refactoring of nannou's internals aimed at improving modularity, compile times, API documentation, reducing maintenance burden and improving nannou's suitability in headless and embedded applications.

# `nannou_core`

nannou's core abstractions.

This crate aims to be a stripped-down foundation for nannou projects that don't require
windowing or wgpu graphics. These might include:

- Headless applications, e.g. for LASER or lighting control.
- Embedded applications, e.g. driving motors or LEDs in an art installation.
- `rust-gpu` shaders which have strict requirements beyond the limitations of `no_std`.
- Hot-loaded dynamic libraries that benefit from faster compilation times.

The crate includes nannou's color, math, geometry and noise abstractions without the deep stack
of crates required to establish an event loop, interoperate with wgpu, etc. Another way of
describing this crate might be "nannou without the I/O".

**Crate `[features]`**

The primary feature of this crate is support for `#![no_std]`. This means we can use the crate
for embedded applications and in some cases rust-gpu shaders.

By default, the `std` feature is enabled. For compatibility with a `#![no_std]` environment be
sure to disable default features (i.e. `default-features = false`) and enable the `libm`
feature. The `libm` feature provides some core functionality required by crates

- `std`: Enabled by default, enables the Rust std library. One of the primary features of this
  crate is support for `#![no_std]`. This means we can use the crate for embedded applications
  and in some cases rust-gpu shaders. For compatibility with a `#![no_std]` environment be sure
  to disable default features (i.e. `default-features = false`) and enable the `libm` feature.
- `libm`: provides some core math support in the case that `std` is not enabled. This feature
  must be enabled if `std` is disabled.
- `serde`: enables the associated serde serialization/deserialization features in `glam`,
  `palette` and `rand`.

Closes #752.
Related to #209, #387, #178.

# `cgmath` -> `glam`

Previously, nannou's linear algebra foundation was provided by `cgmath`, however this PR switches to using the `glam` crate. Motivations for this change include:

- A simpler API, reminiscent of `glm` though more rusty.
- Easier to navigate documentation.
- Super fast compile times. https://github.com/bitshifter/mathbench-rs#build-times
- Support for the `rust-gpu` `spirv-unknown-unknown` target, i.e. writing shaders in Rust.
- Support for `#![no_std]` environments such as embedded, rust-gpu.
- `serde` feature for (de)serialization support.
- `rand` feature providing `rand` crate distribution implementations.

As a part of this transition, the `nannou::geom::vector` module has been removed in favour of leaning on `glam`'s `Vec2/3/4` types. For the most part, both APIs are very similar, bar a few key differences:

- `glam`'s `Vec*` types are opaque and cannot be constructed directly from its fields. In turn a constructor should be used instead, e.g. `Vector2 { x, y }` -> `vec2(x, y)` or `[x, y].into()`.
- The `glam` `magnitude` equivalent method is called `length`
- `glam` does not provide a `Vec*::with_length` method. Instead, users can use `v.normalize() * length` to the same effect.

It should be noted that the `nalgebra-glm` crate was also a close consideration, however after some discussion in the nannou-dev matrix channel and the EmbarkStudios discord I lent towards `glam` primarily for its much faster compile times and use within the new `spirv_std` crate as the standard math library for working with `rust-gpu` shaders.

Related to #387, #178.

# `Draw<S>` becomes `Draw`

This PR removes the generic scalar type parameter on the `Draw` API and related items in the hope of both improving compile times and simplifying the API a little. The only use case for this generic parameter was if a user wanted to instantiate their own `Draw` instance that used `f64` as the scalar type, rather than the default `f32`. However, the `draw` API's render backend has only ever supported using the default `f32` scalar type, so using `f64` would also have required that the user write their own rendering backend - a feat that I doubt anyone would have taken on without getting in touch about making the existing backend more generic.

The removal of this type parameter should hopefully make the draw documentation a little easier to grok, and also remove a lot of the previously required monomorphisation around the draw API and in turn improve re-compilation of downstream crates at least a bit. That said, beware I'm yet to setup any proper compile-time bench-marking for nannou.

# Removing `geom::Graph`

The `geom::Graph` abstraction has been seemingly unused since it was added. Originally I thought it might make for a useful dynamic way of composing more complex geometry together. I never ended up finding a particular use for it, and I feel it's likely even less useful now that the `Draw` API allows for applying arbitrary transformations (scaling, rotation, translation). Please let me know here if you do use it or foresee yourself using it along with your use-case and we can consider re-adding it.

# TODO

- [x] Update downstream crates and examples for breaking changes.
- [x] Update tutorials for API changes.
- [x] Update changelog.
- [x] Add job to CI for testing `#![no_std]` build of `nannou_core`.
- [x] Add publish step for `nannou_core`.

## Future Work

- [ ] Move `noise` module into `nannou_core` - see https://github.com/Razaekel/noise-rs/issues/251
- [ ] Move `text` module into `nannou_core` - blocked on working out what to do about unmaintained `rusttype` crate.


nannou/src/draw/mesh/mod.rs
```diff
@@ -181,75 +176,72 @@ where
...
     }
 }
 
-impl<S> mesh::GetVertex<u32> for Mesh<S>
-where
-    S: Clone,
-{
-    type Vertex = Vertex<S>;
+impl mesh::GetVertex<u32> for Mesh {
+    type Vertex = Vertex;
```
danwilhelm, line 200: Unnecessary

nannou/src/draw/mesh/mod.rs
```diff
@@ -181,75 +176,72 @@ where
...
         mesh::WithTexCoords::get_vertex(&self.mesh, index)
     }
 }
 
-impl<S> mesh::Points for Mesh<S> {
-    type Point = vertex::Point<S>;
-    type Points = Points<S>;
+impl mesh::Points for Mesh {
+    type Point = vertex::Point;
+    type Points = Points;
```
danwilhelm, line 208: Unnecessary

nannou/src/draw/mesh/mod.rs
```diff
@@ -181,75 +176,72 @@ where
...
     fn points(&self) -> &Self::Points {
         self.mesh.points()
     }
 }
 
-impl<S> mesh::Indices for Mesh<S> {
+impl mesh::Indices for Mesh {
     type Index = u32;
     type Indices = Indices;
```
danwilhelm, line 216: Unnecessary

nannou/src/draw/mesh/mod.rs
```diff
@@ -181,75 +176,72 @@ where
...
     fn indices(&self) -> &Self::Indices {
         self.mesh.indices()
     }
 }
 
-impl<S> mesh::Colors for Mesh<S> {
+impl mesh::Colors for Mesh {
     type Color = vertex::Color;
     type Colors = Colors;
```
danwilhelm, line 224: Unnecessary

nannou/src/draw/mesh/mod.rs
```diff
@@ -181,75 +176,72 @@ where
...
         self.mesh.colors()
     }
 }
 
-impl<S> mesh::TexCoords for Mesh<S> {
-    type TexCoord = geom::Point2<S>;
-    type TexCoords = TexCoords<S>;
+impl mesh::TexCoords for Mesh {
+    type TexCoord = geom::Point2;
+    type TexCoords = TexCoords;
```
danwilhelm, line 232: Unnecessary

nannou/src/draw/mod.rs
```diff
@@ -209,195 +198,186 @@ where
...
 
     /// Specify the orientation along each axis with the given **Vector** of degrees.
-    pub fn degrees(&self, v: geom::Vector3<S>) -> Self {
-        self.radians(geom::vec3(
-            deg_to_rad(v.x),
-            deg_to_rad(v.y),
-            deg_to_rad(v.z),
-        ))
+    pub fn degrees(&self, v: Vec3) -> Self {
+        self.radians(vec3(deg_to_rad(v.x), deg_to_rad(v.y), deg_to_rad(v.z)))
```
danwilhelm, line 313: `glam` uses this for their tests, kind of an interesting idea that doesn't involve `math`:

```
pub trait Deg {
    fn to_radians(self) -> Self;
}
      
impl Deg for f32 {
    fn to_radians(self) -> f32 {
        f32::to_radians(self)
    }
}

impl Deg for f64 {
    fn to_radians(self) -> f64 {
        f32::to_radians(self)
    }
}
```

nannou/src/draw/mesh/mod.rs
mitchmindtree: Thanks for taking a look at the PR!

These `type A = B;` lines are not normal type aliases. They are "associated types" of the trait within which they're nested. So despite it looking like a type alias that says `type Foo = Foo;` it's actually part of the implementation of the trait.

In this case it says "the associated type `Vertex` of the implementation of `GetVertex<u32>` for `Mesh` is the type `Vertex`", where the trait's associated type `Vertex` just happens to share the same name as `Vertex` type from the outer scope.

The trait having an associated type like this is useful when writing generic code. By saying `T: GetVertex` in a generic type or function, you can then also talk about `T::Vertex` without having to declare a new generic type parameter for the vertex type. I think the `Iterator` trait's associated type `Item` is probably the most prevalent example.

Going to resolve these, but let me know if you meant something else and I'm missing something!

danwilhelm: Thanks for the explanation, I should have researched it more! I was thrown off by the X = X and initially thought it was a potential "replace all" situation.

nannou/src/draw/mod.rs
mitchmindtree: Yeah it's a pity they don't expose these from the crate! I guess the tricky thing is that using traits like this would begin to conflict if you had more than 2, e.g. if `trait Deg` and `trait Turns` both declare a `to_radians` method and both are implemented for `f32`, you'd need to distinguish which you're talking about at the call-site with `Deg::to_radians(f)` rather than being able to do `f.to_radians()`. Maybe that's not too bad though.

I did add a `ConvertAngle` trait to the `math` mod which provides these functions as methods, so you could do `f.deg_to_rad()`, `f.rad_to_turns()`, etc, but I wasn't sure if it was worth adding it to the prelude if we already have these in the prelude as functions.

danwilhelm: True! Feel free to resolve, this was moreso just a remark 😂 It's really difficult to find anything out of ordinary in your revisions!

PR #759, opened 2021-06-21
Add rand to update dependencies in changelog.md

sidwellr:


mitchmindtree:
Good catch, thanks for the PR!

PR #771, opened 2021-07-15
Update glam to 0.17

dzil123:
See changelog [here](https://github.com/bitshifter/glam-rs/blob/1da7d6459c5f9890275c7fc1871e6c33c825b6b9/CHANGELOG.md).

I need this for the Add and Sub implementations on scalar types, eg Vec2 + f32


mitchmindtree:
Thanks!

PR #773, opened 2021-07-18
Add disable drag-and-drop method for cpal+winit Windows compatibility

danwilhelm:
Fixes #652 
Fixes #483 

**Rationale**
Currently, Windows audio examples crash, because on Windows [`winit` drag and drop requires multi-threaded COM](https://github.com/rust-windowing/winit/issues/1185). This interferes with other crates such as `cpal` and `nannou_audio` on the same thread. This PR allows Windows users to use `nannou_audio` easily on the same thread by disabling `winit` drag and drop. Unfortunately, this solution will not work for Windows users who want both audio and drag and drop.

**Alternatives**
The most common alternative is [spawning `cpal` on another thread](https://github.com/RustAudio/rodio/issues/214#issuecomment-643640535). Other libraries have taken this approach [such as `ruffle`](https://github.com/ruffle-rs/ruffle/blob/master/desktop/src/audio.rs#L77). However, upon comparison it does not  seem as feasible for `nannou`, which attempts to wrap `cpal` much more comprehensively. In my opinion, `nannou`'s existing `Device`/`Host` API would have to be restructured to either allow batching of device queries (and spawn new `cpal` threads each time), or have a continual `cpal` thread with messaging to query devices (which seems overly complex).

**Changes**
- Disables drag and drop in audio examples.
    - **_Alternatively, drag and drop could be disabled by default on Windows -- would this be preferred?_**
- Adds new `WindowBuilder` method `windowsos_drag_and_drop(bool)`. 
    - I prefixed this with `windowsos_`, since `windows_` might be confused with a regular window.
    - The method does nothing on non-Windows OS.
- Adds `stream.play()` to each stream upon initialization.
    - This is required for the examples to work on Windows, since it [allows the callbacks to be called](https://github.com/RustAudio/cpal/issues/516#issuecomment-745238626).

PR #774, opened 2021-07-20
Add community tutorials, disable warnings in conditional compilation

### LINKED ISSUE #740: WHITE is looks green...
Hi I've been testing out but, seems that the `WHITE`or `Rgb(1.0, 1.0, 1.0)` appears to be green. Screenshot attached.

```rust
use nannou::prelude::*;
struct Model {}
fn main() {
    nannou::app(model).update(update).simple_window(view).run();
}
fn model(_app: &App) -> Model {
    Model {}
}
fn update(_app: &App, _model: &mut Model, _update: Update) {}
fn view(app: &App, _model: &Model, frame: Frame) {
    let draw = app.draw();
    let white = Rgb::new(1.0, 1.0, 1.0);
    draw.ellipse().x_y(-60.0, 0.0).w_h(50.0, 50.0).color(WHITE);
    draw.ellipse().x_y(60.0, 0.0).w_h(50.0, 50.0).color(white);
}
```
Screen Shot 2021-05-17 at 6 32 21 PM

I'm using Mac 11.2.3 (M1), rustc 1.50.0, nannaou 0.16.0.

danwilhelm:
- Add community tutorials section to guide.
- Disable conditional compilation warning (to resolve unused variable warning on non-Windows from #773).
- Update changelog.

### DIFF
--- a/guide/src/welcome.md
+++ b/guide/src/welcome.md
@@ -31,6 +31,8 @@ A suite of tutorials demonstrating how to do different things with Nannou. For
 example, "How do I output sounds?", "How do I draw shapes?", "How can I connect
 to my laser?"
 
+> You can find more tutorials by the community [here](/community_tutorials.md).
+
 ### [Developer Reference](/developer_reference.md)
 
 Learn more about the design philosophy behind Nannou, how the project is
--- a/README.md
+++ b/README.md
@@ -36,6 +36,7 @@ It is still early days and there is a lot of work to be done. Feel free to help 
   - [**Create A Project**](https://www.guide.nannou.cc/getting_started/create_a_project.html)
   - [**Upgrading to a New Release**](https://guide.nannou.cc/getting_started/upgrading.html)
 - [**Tutorials**](https://www.guide.nannou.cc/tutorials.html)
+- [**Community Tutorials**](https://www.guide.nannou.cc/community_tutorials.html)
 - [**Developer Reference**](https://www.guide.nannou.cc/developer_reference.html)
 - [**API Reference**](https://www.guide.nannou.cc/api_reference.html)
 - [**Showcases**](https://www.guide.nannou.cc/showcases.html)
--- a/guide/src/SUMMARY.md
+++ b/guide/src/SUMMARY.md
@@ -20,6 +20,7 @@
     - [Draw - Drawing Images](./tutorials/draw/drawing-images.md)
     - [OSC - Introduction](./tutorials/osc/osc-introduction.md)
     - [OSC - Sending OSC](./tutorials/osc/osc-sender.md)
+- [Community Tutorials](./community_tutorials.md)
 - [Contributing](./contributing.md)
     - [About the Codebase](./contributing/about-the-codebase.md)
     - [PR Checklist](./contributing/pr-checklist.md)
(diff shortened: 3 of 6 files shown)


guide/src/welcome.md
```diff
@@ -31,6 +31,8 @@ A suite of tutorials demonstrating how to do different things with Nannou. For
 example, "How do I output sounds?", "How do I draw shapes?", "How can I connect
 to my laser?"
 
+	- [Community Tutorials](/community_tutorials.md)
```
mitchmindtree, line 34: This will end up formatted a little strangely - maybe instead we can add another sentence along the lines of

> You can find more tutorials by the community [here](./community_tutorials.md).

nannou/src/draw/renderer/mod.rs
```diff
@@ -974,6 +974,18 @@ impl Renderer {
...
+        // Ensure background is cleared on first frame if no clear specified.
+        // A previously cleared frame is required for wgpu::LoadOp::Load.
+        if frame.nth() == 0 {
+            let bg_color = draw.state.borrow().background_color;
+            if bg_color.is_none() {
+                if let Ok(mut state) = draw.state.try_borrow_mut() {
+                    state.background_color = Some(Default::default());
+                }
+            }
+        }
```
mitchmindtree, line 987: This might sound a little crazy, but I think we should also consider the case where a user might have a transparent window - is it still possible with a setup like this? I guess it might work still as long as `background_color` is set to a zero-alpha value?

mitchmindtree (commented):
This is a great idea! Nice one @danwilhelm :)

nannou/src/draw/renderer/mod.rs
danwilhelm: I verified that if the window is transparent, clearing it with alpha=1. results in a colored background, while alpha=0. results in transparent. So I'll move this code into `to_frame()` and set the alpha based on whether the window is transparent.

danwilhelm: If you resize the window or restore it later, I'm unsure whether the frame needs to be re-cleared. I noticed the frame &  window counters just continually increase no matter what, so they wouldn't be useful in that situation.

danwilhelm:
I believe there is something wrong with the automated checks -- cannot compile `hotglsl`?

- Moved the background clear into another PR, since I realized there are additional cases to handle. 
- Updated the community tutorial reference as suggested to a `>` rather than a `-`.

nannou/src/draw/renderer/mod.rs
mitchmindtree: > If you resize the window or restore it later, I'm unsure whether the frame needs to be re-cleared

Yeah I think clearing on the first frame is probably enough, it's not super clear that there's a single correct behaviour for most users when it comes to resizing (e.g. we might clear over a long-running additive sketch), so it might be best to leave that case up to the user.

mitchmindtree:
> I believe there is something wrong with the automated checks -- cannot compile hotglsl?

Ahhh looks like there might have been a breaking change upstream in `notify`, I'll do some digging. I think this is fine to merge :)

nannou/src/draw/renderer/mod.rs
danwilhelm: I noticed it always loses the old frame data upon resize, because it always allocates a new frame buffer. So I concluded it does need clearing on the first frame of each new framebuffer.

This seems consistent with the recent user-reported issue on Matrix, who said the fullscreen transition happened on about the tenth frame. It only showed uninitialized contents (weird color tint) when it moved to fullscreen.

PR #775, opened 2021-07-25
Clear background when window invalidated

### LINKED ISSUE #740: WHITE is looks green...

danwilhelm:
Fixes #740 (assumed -- cannot replicate the issue)

Nannou does not guarantee that a new framebuffer is initially cleared,
sometimes causing uninitialized buffer contents to be displayed.

1. Adds `is_invalidated` and `clear_color` attributes to `Window`.
2. Allows a custom clear color via `WindowBuilder::clear_color()`.

**Potential disadvantages:**
1. An extraneous clear will occur if the user explicitly clears the background 
in `view()`. Without exposing more internals to the user, this does not seem 
preventable. Also, these clears occur rarely -- only when the window is invalidated.


mitchmindtree (commented):
Nice one @danwilhelm!

> An extraneous clear will occur if the user explicitly clears the background in view().

I think this is easily worth the win! I think the new `WindowBuilder::clear_color` method is a much nicer way for users to clear only the first frame (rather than checking for `frame.nth() == 0`). Perhaps at some point we can also add a `SketchBuilder`/`app::Builder` method for setting this to allow for users of `nannou::sketch` to also take advantage of this.

nannou/src/window.rs
```diff
@@ -241,6 +243,8 @@ pub struct Window {
     pub(crate) frame_count: u64,
     pub(crate) user_functions: UserFunctions,
     pub(crate) tracked_state: TrackedState,
+    pub(crate) is_invalidated: bool,
```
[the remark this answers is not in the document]
danwilhelm: I named it this because "invalidate" is often used in GUIs to indicate that a window must be redrawn (e.g. [Windows](https://docs.microsoft.com/en-us/windows/win32/gdi/invalidating-the-client-area)). I'll add a comment for now, but if you prefer the other name let me know and I'll switch it!

danwilhelm:
> I think the new `WindowBuilder::clear_color` method is a much nicer way for users to clear only the first frame.

And if `clear_color` is not run, it will auto-clear to black (or transparent). This is to make sure users aren't confused if the background is uninitialized.

Example of a custom clear color (using redfrik's [cuboids demo](https://fredrikolofsson.com/software/nannou/cuboids/)):

https://user-images.githubusercontent.com/6137185/127267646-52f4189b-52c3-43b0-aa0a-132cf65862db.mp4

nannou/src/app.rs
```diff
@@ -1175,6 +1175,12 @@ fn run_loop<M, E>(
                             window_rect,
                         );
 
+                        // Clear the raw frame immediately once the window is invalidated
+                        if window.is_invalidated && frame_data.is_some() {
+                            let data = frame_data.as_ref().expect("missing `frame_data`");
+                            raw_frame.clear(&data.render.texture_view(), window.clear_color);
+                        }
```
dzil123, line 1182: Maybe you can use an `if let` to get rid of the `.expect()`?

```suggestion
                        if window.is_invalidated {
                            if let Some(data) = frame_data {
                                raw_frame.clear(&data.render.texture_view(), window.clear_color);
                            }
                        }
```

danwilhelm: Thanks for taking a look! The downside is that it adds nesting -- which I typically try to avoid. But, I too suspect that it might be more canonical. So, I just tested & re-committed it with the change.

Issue #379, opened 2019-08-07
Export to SVG file


sethboyles:
An "Export to SVG file" would be something handy for users who would like to take images from their nannou sketches and work with them externally (e.g. printing, laser cutting, axidraw, etc).  

Since [lyon](https://github.com/nical/lyon) is in the process of being added to nannou (see https://github.com/nannou-org/nannou/pull/375), it'd probably be best to look at how to convert the results from lyon to SVG.  

As noted in the lyon README:

```
How do I render an SVG file with lyon?
Lyon is not an SVG renderer. For now lyon mainly provides primitives to tessellate complex path fills and strokes in a way that is convenient to use with GPU APIs such as gfx-rs, glium, OpenGL, D3D, etc. How the tessellated geometry is rendered is completely up to the user of this crate.
```

sethboyles:
For reference: Processing and OpenFrameworks have solutions for this:
* [Processing](https://processing.org/reference/libraries/svg/index.html)
* [OpenFrameworks](h ttps://openframeworks.cc/documentation/graphics/ofGraphics/#!show_ofBeginSaveScreenAsSVG)

mitchmindtree:
Thanks for opening this @sethboyles!

If you're mostly interested in drawing the contents of a `Draw` instance (the type often yielded by `app.draw` in nannou examples), one quick approach that might be worth looking into is the [`Draw::triangles`](https://docs.rs/nannou/0.9.0/nannou/draw/struct.Draw.html#method.triangles) method.

> Produce an iterator yielding all triangles from the inner mesh transformed via the inner geometry graph.
> 
> Triangles are yielded in depth-first-order of the geometry graph nodes from which they are produced.
> 
> Note: If there are any Drawings in progress, these will first be drained and completed before any vertices are yielded.

Each triangle contains the color and position data for each vertex. Perhaps these triangles can be used alongside the [svg crate](https://crates.io/crates/svg) somehow to produce the SVG? It wouldn't produce the most optimal SVG (e.g. tessellation of all the primitives would already be performed) but it might be a good-enough work around in the meantime depending on the use case?

I agree it would be nicer to have a proper solution in the future. One thought that crosses my mind is, rather than tessellating each shape as they're drawn, we could store them as a list of commands in the `Draw` itself. Then this list of commands could either be rendered to a more efficient SVG path representation, or tessellated for submission to a shader, etc.

duncangeere:
Would love to see this too.

[...]

PR #778, opened 2021-07-28
Refactor `draw::renderer::RenderPrimitive` to enable alternative renderers

dzil123:
Creates a `PrimitiveRenderer` trait to allow separation of the lyon tesselation and wgpu rendering from the general preparation and finalization of the different primitives.

Allows alternative renderers to avoid duplicating existing code that normalizes the nannou primitives into simpler rendering primitives.

Preparation for render+export to SVG: #379


dzil123:
This PR is at the point where it can render the `draw.rs` example. Here is a sketch that renders the window to svg on mouse click: https://gist.github.com/dzil123/f52f4d554c570188ddcf73483a2dc292.

@mitchmindtree, would you mind taking a look?

fenjalien:
Hi I've tried making my own program to generate a Sierpinski triangle SVG file and it works really well! One thing I didn't realise is it only renders whats been drawn by the `Draw` struct in the `view` method.

See here: https://gist.github.com/fenjalien/bc9fa8ca765470308ea311ea68fb499a

dzil123:
Thanks for checking it out!

I see what you mean about it rendering only whats drawn by the `Draw` that single `view` call. Nannou doesn't remember all the draw commands from previous frames, so fixing this isn't trivial. It just loads last frame's rendered texture and draws on top of it if `draw.background()` wasn't set.

I don't know if it would be practical to store previous draw commands indefinitely because it can grow without bounds if not cleared with a background. Maybe it could be opt-in? It could also be useful to redraw everything when the render texture is lost due to a window resize.

danwilhelm:
It probably does make sense to constrain the SVG render to the commands within a single `Draw` instance -- without incorporating texture contents.

There are many situations that would have to be handled otherwise, e.g.:
- `frame.clear()` (bypassing `app.draw()`),
- Multiple `Draw` instances,
- Use of `wgpu` (i.e. shaders, blending, etc.), and
- Drawing on top of an existing texture as mentioned.

mmansion:
Great work on this so far! SVG rendering would be an excellent enhancement. Looking to seeing it integrated soon.

akonneker:
I wanted to leave a nudge here. I'm currently setting up to use nannou to generate some input files for a laser marking system. I would like to export svgs... and so this PR would be immensely helpful. I'm currently working off my own fork of this PR in the meantime. 

If the underlying abstractions are still sound and this has potential to be merged with a reduced scope (e.g. ignoring meshes and stroke end cap stuff), I may be able to sponsor someone to work on it. I've not been involved in the nannou project, so I'm not sure how many active contributors there are, or if that is a possible or desirable arrangement.

PR #780, opened 2021-07-31
Received character keyboard event

### LINKED ISSUE #480: Add in an event to retrieve keyboard events as a string
Currently there is no way to easily get key pressed events as a String. Conrod has this feature for example.

gmunk:
Adding a "simplified" event, as well as a handler function for receiving the character value of a pressed keyboard key.

This PR resolves #480.


mitchmindtree:
Thanks for this!

PR #782, opened 2021-08-05
Don't update model in LoopMode::Wait when there were no events

### LINKED ISSUE #781: LoopMode::Wait not working on macOS
Nannou version: `0.17.1`
macOS version: Big Sur (11.4)

After I `app.set_loop_mode(LoopMode::wait());`, I expect my view function to be called only on input events. However, it seems to be called for each frame at 60fps.

Is this working on other platforms? I am happy to take a crack at working on this but I might need some guidance.

Here are some events that are flowing through the `winit` event loop:
```
...
NewEvents(WaitCancelled { start: Instant { t: 1006573215416347 }, requested_resume: None })
MainEventsCleared
RedrawRequested(WindowId(Id(140410039122816)))
RedrawEventsCleared
NewEvents(WaitCancelled { start: Instant { t: 1006573258088858 }, requested_resume: None })
MainEventsCleared
RedrawRequested(WindowId(Id(140410039122816)))
RedrawEventsCleared
...
```

Taking a quick look at `app.rs`, it seems like `MainEventsCleared` always results in an invocation of the user's view function, unless using `LoopMode::NTimes` where all number of renders has exceeded "n". I am still having trouble understanding how the loop "knows" to render at 60fps, or put another way, why the `winit` `ControlFlow::Wait` is being interrupted.

zgotsch:
Pending some clarity from the `winit` maintainers about whether wait interruptions without events are allowed, this is an idea for a fix for `LoopMode::Wait` if they are.

See https://github.com/rust-windowing/winit/issues/1985 for more info.
Fixes #781

### DIFF
--- a/nannou/src/app.rs
+++ b/nannou/src/app.rs
@@ -190,6 +190,7 @@ struct LoopState {
     loop_start: Instant,
     last_update: Instant,
     total_updates: u64,
+    events_since_wakeup: usize,
 }
 
 /// The mode in which the **App** is currently running the event loop and emitting `Update` events.
@@ -1074,6 +1075,7 @@ fn run_loop<M, E>(
         loop_start,
         last_update: loop_start,
         total_updates: 0,
+        events_since_wakeup: 0,
     };
 
     // Run the event loop.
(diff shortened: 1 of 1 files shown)


mitchmindtree:
Thanks @zgotsch, looks good to me!

PR #783, opened 2021-08-08
Specifying an initial, default LoopMode.

### LINKED ISSUE #528: Allow for specifying default initial `LoopMode` in app and sketch builder
This would allow for specifying alternative loop modes in sketches without the need to do so in the `view` function, the behaviour of which can be a bit ambiguous.

gmunk:
Created two methods, for `SketchBuilder` and `Builder` respectively, which allow for specifying an initial LoopMode. Updated the relevant example to showcase the new functionality and fixed an unrelated, lingering code warning.

This resolves #528.

### DIFF
--- a/examples/nannou_basics/loop_mode.rs
+++ b/examples/nannou_basics/loop_mode.rs
@@ -10,14 +10,16 @@
 use nannou::prelude::*;
 
 fn main() {
-    nannou::app(model).update(update).run();
+    // Start in `Wait` mode. In other words, don't keep looping, just wait for events.
+    nannou::app(model)
+        .loop_mode(LoopMode::Wait)
+        .update(update)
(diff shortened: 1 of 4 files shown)


danwilhelm:
The implementation looks great! I did notice some other examples that may have to be updated (in `/examples/ui/conrod/` and in ` /generative_design/`). 

I assume the original function should be left in, both for backwards compatibility and for things like this (from `generative_design/shape/p_2_2_3_01.rs`):

```rust
            if model.freeze {
                app.set_loop_mode(LoopMode::loop_once());
            } else {
                app.set_loop_mode(LoopMode::RefreshSync);
            }
```

mitchmindtree:
Looks great, thanks for this!

PR #784, opened 2021-08-08
Add headless feature

fenjalien:
Allows nannou to run apps without having to use `winit` to create a window. 
It Builds off of @dzil123 's proof of concept branch (https://github.com/dzil123/nannou/tree/headless) and provides a similar API to non-headless.

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -18,15 +18,20 @@ audrey = "0.3"
 futures = "0.3"
 hotglsl = { git = "https://github.com/nannou-org/hotglsl", branch = "master" }
 hrtf = "0.2"
-nannou = { version ="0.17.0", path = "../nannou" }
-nannou_audio = { version ="0.17.0", path = "../nannou_audio" }
-nannou_conrod = { version ="0.17.0", path = "../nannou_conrod" }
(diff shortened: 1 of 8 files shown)


dzil123:
This is a good start! I would look into the idea of merging headless::App and normal App, so that sketches can run headless without having to change the App type in function signatures. This would also help with deduplicating the Builder and Fn types.

My suggestion is to change `window:Id` to be an enum over `winit::window::WindowId` and some marker for a "headless" window (probably a usize because that's essentially what a winit WindowId is). This would also pave the way for headless windows to coexist with regular windows.

fenjalien:
I've started implementing your suggestion @dzil123 but run into a bit of a roadblock. What should the id be a key to?

I've tried using it as a key to `App`'s `window` hashmap attribute but that requires a second enum over the two types of windows. To me that would be annoying to pattern match against a type I'm not even using. Having separate hashmaps still runs into the same issue due to `App`'s `window` method. 
A solution would be to use separate methods for each type but that would remove the idea of the id to begin with.

Any ideas on this? I think we should try to find as many options as possible.

Also what should non-headless windows be called in the enum?

PR #785, opened 2021-08-09
Remove Windows drag-and-drop method

danwilhelm:
Removes Windows-specific option to disable drag-and-drop.

This function was originally included to allow `cpal` and `winit` to co-exist on the same thread. However, [`cpal` now uses a compatible threading model](https://github.com/RustAudio/cpal/pull/597).


JoshuaBatty:
Great thanks, @danwilhelm !

PR #787, opened 2021-08-20
Fixes Looping request if LoopMode::Wait

nmpribeiro:
This is a tentative fix to the LoopMode::Wait behaviour.


nmpribeiro:
Todo: figure out how to only trigger re-draw/loop on events inside the window.

mitchmindtree:
Thanks a lot for opening this @nmpribeiro - I think this has since been addressed by #782, #818 and could be improved a little by #819. Going to close this for now, but feel free to re-open if you're still running into issues with `LoopMode::Wait`.

PR #795, opened 2021-09-21
Update `wgpu` to version `0.10`

mitchmindtree:
The changes for wgpu are listed here:

https://github.com/gfx-rs/wgpu/blob/master/CHANGELOG.md#v010-2021-08-18

## ~`SwapChain`~ `Surface`

Most notably, the concept of the `SwapChain` has been removed by merging it into the `Surface` :tada: This simplifies some internals quite a bit, and hopefully makes the wgpu parts of nannou just a little more approachable.

## WGSL Shaders

All `.glsl` and `.spv` shaders have been replaced with `.wgsl` shaders. For most I used `naga` to translate the original `.glsl` shader to `.wgsl` automatically, and then cleaned up the `.wgsl`. There was one shader that required generating the `.wgsl` from the `.spv` instead, I think perhaps because the `.glsl` input on naga didn't yet know about one of the built-in glsl functions. I believe it was `textureSize`? I don't quite remember, but all appears to be working nicely now.

---

## TODO

This PR is mostly complete, though blocked on a couple more steps:

- [x] Update naga-translated `draw::renderer` wgsl shaders to be a bit more human-friendly.
- [x] Publish a new `conrod_wgpu` version.
- [ ] Get `wgpu_compute_shader.rs` and `isf_demo.rs` examples working on Intel.
      Currently there's a bug where on intel, only one command buffer can be submitted per frame.
      https://github.com/gfx-rs/wgpu/issues/1672
      I've updated the texture constructors to account for this, however the `wgpu_compute_shader.rs` performs a standalone command buffer submission in the `update` before another command buffer is submitted for drawing to the window's surface in the `view`. Not sure how to go about working around this, as the point of the example is to demonstrate how to process some data on the GPU before using it again on the CPU. The `isf_demo.rs` also freezes/crashes, I believe due to the initial submission of a command buffer that creates the uniform buffer. For the record both examples should be valid as they are, the bug appears to be an issue with Intel's Vulkan driver (on Linux at least).


nannou_wgpu/src/texture/image.rs
```diff
@@ -564,14 +614,71 @@ where
...
+    // of using `Queue::write_texture` which schedules the write for the next call to
+    // `Queue::submit`. This is to avoid an Intel driver bug where submitting more than one command
+    // buffer per frame appears to be causing issues:
+    // https://github.com/gfx-rs/wgpu/issues/1672#issuecomment-917510810
+    //
+    // While this likely means consuming more RAM, it also likely results in slightly better
+    // performance due to reducing the number of command buffers submitted.
+    //
+    // Users can still use `encode_load_texture_array_from_image_buffers` directly if they wish.
+    let capacity = bytes_per_row as usize * height as usize * array_layers as usize;
```
danwilhelm, line 660: `bytes_per_row` => `padded_bytes_per_row`

nannou_wgpu/src/texture/image.rs
```diff
@@ -564,14 +614,71 @@ where
...
+            data.extend((0..padding).map(|_| 0u8));
+        }
+    }
+
+    // Copy into the entire texture.
+    let image_copy_texture = wgpu::ImageCopyTexture {
+        texture: texture.inner(),
+        mip_level: 0,
+        origin: wgpu::Origin3d::ZERO,
+        // TODO: Maybe we shouldn't assume this?
```
danwilhelm, line 675: A quick way not to assume it might be `let image_copy_texture = texture.as_image_copy();`

nannou_wgpu/src/texture/image.rs
danwilhelm, line 660: `bytes_per_row` => `padded_bytes_per_row`

nannou_wgpu/src/texture/image.rs
mitchmindtree: Agh of course, I forgot about that method, cheers!

mitchmindtree:
I think this is just about ready to go!

As the Intel Vulkan driver bug is a little out of our hands, and as it only occurs when attempting to submit more than a single command buffer per frame, I think it's fine to go ahead and land this as long as we open a nannou issue to track upstream progress on the Intel iGPU issues upstream. Hopefully this is eventually patched by Intel, or we can work out a more general workaround in collaboration with wgpu in a patch release.

PR #797, opened 2021-09-26
Remove daggy (unused dependency)

dlight:
Since daggy isn't used anywhere in the crate, this just bumps the version number in `nannou/Cargo.toml`.


dlight:
Another option is to simply not depend on daggy anymore, and let people depend on it themselves. That way people can depend on the latest version of the crate without needing a new nannou release.

dlight:
@mitchmindtree done!

Issue #799, opened 2021-10-04
Drawing stroke is inaccurate at high scale


Frojdholm:
Using `nannou` version 0.17.1 when drawing using a high `scale` and `no_fill()` the strokes are not accurate. I'm on Windows 10 using `rustc 1.55.0 (c8dfcfe04 2021-09-06)`.

See the following example

```rust
use nannou::prelude::*;

fn main() {
    nannou::sketch(view).run();
}

fn view(app: &App, frame: Frame) {
    // Inaccurate stroke
    let draw = app.draw().scale(800.0);
    draw.rect().w_h(0.03, 0.03).stroke_color(SNOW).no_fill().stroke_weight(1.0 / 800.0);

    // Fill works
    // let draw = app.draw().scale(800.0);
    // draw.rect().w_h(0.03, 0.03).color(SNOW);

    // Not using scale works
    // let draw = app.draw();
    // draw.rect().w_h(0.03 * 800.0, 0.03 * 800.0).stroke_color(SNOW).no_fill().stroke_weight(1.0);

    draw.to_frame(app, &frame).unwrap();
}
```
Which produces this image:
![inaccurate_stroke](https://user-images.githubusercontent.com/3251566/135892006-e8e1d76c-edd8-4a58-8581-9e04fe40c4bc.png)

Frojdholm:
After some investigation it looks like it's either a bug or threshold in Lyon causing vertices to be merged when tessellating the stroke with small stroke width.

Since the transform is applied after tessellating the points and stroke width passed to Lyon are very small.

Frojdholm:
A fix is now merged in Lyon and available from 0.17.10.

mitchmindtree:
Closed via #801, thanks again @Frojdholm!

PR #801, opened 2021-10-16
Update lyon to version 0.17

Frojdholm:
The bug in #799 has been fixed in lyon 0.17.10. This PR updates to the latest version of lyon, updating the affected code. However, this restricts the functionality of the `nannou::geom::path::Builder` since much of its interface has been moved to SVG-only builders. In particular functions like `move_to` and `arc` can only be used through a `lyon::path::builder::WithSvg<nannou::geom::path::Builder>` with the lyon-specific types, making them less user-friendly.

It would be possible to store a `WithSvg<Builder` directly in `nannou::geom::path::Builder` and implement the `SvgPathBuilder` trait instead of `PathBuilder` to restore much of the lost functionality. Although, I don't know if the added convenience is worth the added complexity, since the `Builder` API seems to be pretty low-level already. It might be better to just remove the `nannou::geom::path::Builder` entirely and just use the lyon types directly.

TODO:
- [ ] Add to changelog.

### DIFF
--- a/generative_design/data_trees/m_5_1_01.rs
+++ b/generative_design/data_trees/m_5_1_01.rs
@@ -24,6 +24,7 @@
  * 0-9                 : recursion level
  * s                   : save png
  */
+use nannou::lyon;
 use nannou::prelude::*;
 
 fn main() {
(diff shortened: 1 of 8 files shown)


mitchmindtree:
Thanks so much @Frojdholm! This looks great to me.

> It might be better to just remove the nannou::geom::path::Builder entirely and just use the lyon types directly.

Yeah this crosses my mind sometimes too. From memory the idea behind the `nannou::geom::path::Builder` was to try and provide a more nannou-esque API (allowing to use the nannou point types I think was the main one), though I'm unsure if it's worth the upkeep. I'm happy to leave things as they are for now and land this.

Issue #806, opened 2021-11-19
No way to start a path/sub-path with the path Builder


abusch:
It looks like as part of the lyon 0.17 upgrade the `move_to()` method got removed from `geom::path::Builder` (see [this commit](https://github.com/nannou-org/nannou/commit/08139bcf213937e07514bc8dbed8aebf8e7508a1#L100) ).

It seems like the underlying method in lyon got renamed to `begin()` instead, so nannou's builder should probably expose that.

PR #807, opened 2021-11-19
Expose `begin()` method in `geom::path::Builder`

abusch:
As part of the lyon 0.17 upgrade, the `move_to` method got removed from
the builder. This makes it impossible to start a path/subpath without
having to access the underlying lyon Builder.

This commit exposes the `begin()` method in nannou's Builder, which is
what `move_to` got renamed to in lyon.

closes: #806

### DIFF
--- a/nannou/src/geom/path.rs
+++ b/nannou/src/geom/path.rs
@@ -95,6 +95,14 @@ impl Builder {
         lyon::path::builder::Flattened::new(self, tolerance)
     }
 
+    /// Sets the position in preparation for the next sub-path.
+    ///
+    /// If the current sub-path contains edges, this ends the sub-path without closing it.
+    pub fn begin(mut self, to: Point2) -> Self {
+        self.builder.begin(to.to_array().into());
+        self
+    }
+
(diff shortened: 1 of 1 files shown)


nannou/src/geom/path.rs
```diff
@@ -95,6 +95,14 @@ impl Builder {
         lyon::path::builder::Flattened::new(self, tolerance)
     }
 
+    /// Sets the position in preparation for the next sub-path.
+    ///
+    /// If the current sub-path contains edges, this ends the sub-path without closing it.
+    pub fn begin(mut self, to: Point2) -> Self {
```
abusch, line 101: We could also keep the `move_to` name here, to avoid breaking API. Or even having both `move_to` and `begin` and deprecating the former?

mitchmindtree: I'm open to re-adding `move_to` with a deprecation notice if you'd like, but otherwise I'm happy for us to keep a minimal set of methods that follow what lyon itself exposes.

mitchmindtree:
Thanks for catching this and taking the time to submit a PR @abusch!

PR #809, opened 2021-11-27
Fix example link in README

chohner:
File was moved in https://github.com/nannou-org/nannou/commit/25f92acc3067710ec8a2ec0cb5fa869d78b80370


mitchmindtree:
Thanks!

PR #812, opened 2021-12-05
Add GL backend to default backends for better WASM support

Woyten:
To enable WASM support in Nannou we should aim for WPGU's WebGL2 backend first since WebGPU support in browsers is still very experimental.

This PR proposes to add `Backends::GL` to the list of default backends s.t. running a Nannou app in the browser does not fail with `BuildError::NoAvailableAdapter`.


mitchmindtree:
Looks good to me!

PR #814, opened 2021-12-06
Update wgpu and winit to the latest version

Woyten:
This PR explicitly lifts wgpu to 0.11.1 which fixes some driver compatibility regressions and is compatible with winit 0.26.0.
winit 0.26.0 has the `web-sys` feature flag removed which I take as a commitment to builtin / reliable WASM support.

Not entirely sure about it but I think this PR also lifts the `rustflags = ["--cfg", "web_sys_unstable_apis"]` requirement of #715. As a result I propose to merge this PR before #715.

### DIFF
--- a/nannou/Cargo.toml
+++ b/nannou/Cargo.toml
@@ -29,8 +29,8 @@ serde_derive = "1"
 serde_json = "1"
 toml = "0.5"
 walkdir = "2"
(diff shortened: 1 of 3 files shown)


mitchmindtree:
Thanks for this!

PR #815, opened 2021-12-07
Add async twin functions for WASM support

Woyten:
This PR exposes a set of `async` functions that are sufficient to run a Nannou app on WASM. The change is conservative i.e. it only exposes a minimal set of functions and it preserves the behavior of the old non-`async` functions.


mitchmindtree:
This is so damn cool. Exciting to see some async support make its way into nannou!

Woyten:
> This is so damn cool. Exciting to see some async support make its way into nannou!

The cool thing is that it's surprisingly easy!

PR #816, opened 2021-12-11
Set window class for X11 running apps

eirikblekesaune:
When running window managers such `i3`, it is nice to be able to specify rules for certain classes of windows. Currently, `nannou` doesn't set this attribute when making a X11 window, thus the `class` attribute in X11 is set to the window title by default.

This PR adds `nannou` as the window class, so that you can for example in `i3` put this in your config file to open all nannou windows as floating windows:
```
for_window [class="nannou"] floating enabled
```

Since `X` is running on multiple OSes, I have not added any config check for this. This will only have effect on systems actually running `X`.

### DIFF
--- a/nannou/src/window.rs
+++ b/nannou/src/window.rs
@@ -758,11 +758,11 @@ impl<'app> Builder<'app> {
         // Set the class type for X11 if WindowExtUnix trait is compiled in winit
         // (see lines https://docs.rs/winit/0.26.0/src/winit/platform/unix.rs.html#1-7)
         #[cfg(any(
-                target_os = "linux",
-                target_os = "dragonfly",
-                target_os = "freebsd",
-                target_os = "netbsd",
-                target_os = "openbsd"
+            target_os = "linux",
+            target_os = "dragonfly",
+            target_os = "freebsd",
+            target_os = "netbsd",
+            target_os = "openbsd"
         ))]
         {
             use winit::platform::unix::WindowBuilderExtUnix;
--- a/nannou/src/window.rs
+++ b/nannou/src/window.rs
@@ -755,9 +755,11 @@ impl<'app> Builder<'app> {
             }
         }
 
-        // Set the class type for X11
-        use winit::platform::unix::WindowBuilderExtUnix;
-        window = window.with_class("nannou".to_string(), "nannou".to_string());
+        // Set the class type for X11 if running unix family.
+        if cfg!(unix) {
+            use winit::platform::unix::WindowBuilderExtUnix;
+            window = window.with_class("nannou".to_string(), "nannou".to_string());
+        }
 
         // Set default dimensions in the case that none were given.
         let initial_window_size = window
--- a/nannou/src/window.rs
+++ b/nannou/src/window.rs
@@ -755,8 +755,16 @@ impl<'app> Builder<'app> {
             }
         }
 
-        // Set the class type for X11 if running unix family.
-        if cfg!(unix) {
+        // Set the class type for X11 if WindowExtUnix trait is compiled in winit
+        // (see lines https://docs.rs/winit/0.26.0/src/winit/platform/unix.rs.html#1-7)
+        #[cfg(any(
+                target_os = "linux",
+                target_os = "dragonfly",
+                target_os = "freebsd",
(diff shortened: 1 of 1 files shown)


nannou/src/window.rs
```diff
@@ -755,6 +755,10 @@ impl<'app> Builder<'app> {
             }
         }
 
+        // Set the class type for X11
```
Woyten, line 758: I believe you need to put this behind an `cgf` architecture toggle. The test suite should catch this but for some reason it has not been run...

eirikblekesaune: Thanks @Woyten for you review. I considered doing so, but I was unsure which architectures to filter for. Maybe you could help me on this? Which architectures doesn't X run on?
Would it be sufficient to add?:
```
#[cfg(target_family = "unix")]
```

eirikblekesaune: It may be that the `cfg` toggle happens in `winit` (https://docs.rs/winit/latest/src/winit/platform/unix.rs.html#252) although filtered on `#[cfg(feature = "x11")]`, which could explain why the test suite doesn't complain(?)

If that is the case, the only thing that gets compiled in is the `use` statement AFAIK.

I will look into where the `feature="x11"` attribute is set. If it is set as a part of the compilation scope of `nannou`, then I would propose using that to toggle the `cfg` setting.

eirikblekesaune:
I tried to check for feature="x11" specifically, but apparently this feature wasn't set in the compilation scope for `nannou`. If we assume that only `unix` family systems run `X`, this code should work fine.

nannou/src/window.rs
Woyten: > Maybe you could help me on this?

Sure. I expect `if cfg!(unix)` to *not* work since Rust will transform this line to either `if true` or `if false` but it will try to `use winit::platform::unix::WindowBuilderExtUnix` regardless of the architecture. You should probably use
```
#[cfg(target_os = "linux"]
{
  // Your code here
}
```

For an exhaustive list of all target architectures that provide the `WindowBuilderExtUnix` trait have a look at https://docs.rs/winit/0.26.0/src/winit/platform/unix.rs.html#1-7.

eirikblekesaune: Thanks for helping out on this. I have a question regarding:
> it will try to use `winit::platform::unix::WindowBuilderExtUnix` regardless of the architecture.

Just so I understand better: So, when using `use` statement inside the `if` block it actually compiles with it? Whereas using `#[cfg(target_os = "linux")]` avoids it becoming compiled into the binary? Just like an `#ifdef` statement in `C++`?

I do get the feeling that using the same `cfg` statement isn't as robust as using the `feature="x11"` attribute. This is however set in `winit`s `Cargo.toml`. Then, how would you get this attribute; how do you get Cargo features form another crate?

eirikblekesaune: Ahh, looking at  https://github.com/rust-windowing/winit/blob/master/Cargo.toml#L88, I see that the list used in the `unix.rs` file is the same.

Woyten:
@eirikblekesaune Rust offers no official way for a crate to evaluate whether `winit/x11` is set except for the crate `winit` itself. However it's okay to assume that `winit/x11` is enabled since it's configured to be enabled by default and Nannou does not override it.

nannou/src/window.rs
Woyten: > Just so I understand better: So, when using `use` statement inside the `if` block it actually compiles with it? Whereas using `#[cfg(target_os = "linux")]` avoids it becoming compiled into the binary? Just like an `#ifdef` statement in `C++`?

This seems correct!

mitchmindtree:
Thanks for this!

Apologies for the wait on the CI run, github recently made it so that maintainers now need to approve and run workflows first (I guess for security and keep their load lower?). I need to look into whether I can re-enable automated runs again somehow. It looks like this needs a `cargo fmt --all` from the top level, otherwise looks good to me.

Ideally I think we should forward `winit`'s features in nannou too as we have begun to do for `wgpu`, but we can leave that for a future PR.

eirikblekesaune:
Thanks @mitchmindtree ,
I have pushed a formatitng fix now.

Issue #728, opened 2021-04-21
Example programs don't exit by clicking the close button


qingxiang-jia:
**Background:**
Pulled on 4/18/2021, built all examples by running: `cargo build --release --examples`.
OS: ArchLinux
DE: GNOME 40

**Steps:**
1. Run any example (either from terminal or by clicking it).
2. Click the close button.

**Expect:**
The app stops running and the window is closed.

**Actual:**
The window is closed but the app continues to run. This can be checked by running System Monitor or the equivalent commandline tool.

**Note:**
If you run from the terminal, and ctrl+c it, the app will be terminated.

danwilhelm:
Sounds interesting, I don't see any issues about this yet in `winit` or here.

- Could you try `cargo run --release --example simple_window`? Based on this report, when you close the window it closes, but the application hangs in the terminal until you press Ctrl+C?

- There are quite a few changes in the latest master, so I wonder if this problem was introduced since the last release. Perhaps you could try `git checkout v0.15.0` then the same command as above?

mitchmindtree:
I'm also on GNOME and noticed this a day or two ago too! Specifically, pressing `Esc` works fine - it's only pressing the `x` button on the window that seems to result in this behaviour.

If I

1. run the `draw.rs` example
2. attach to the process with `gdb -p <pid>`
3. press the close button and then
4. press ctrl+c on the command line,

gdb is able to catch the interrupt and the state of the program. Here's the backtrace:

```
(gdb) backtrace
#0  0x00007ff39644d3f7 in __pthread_clockjoin_ex () from /nix/store/0qrayfgw779phhbasnrynvyakpczwdkb-glibc-2.32-39/lib/libpthread.so.0
#1  0x00007ff3817f7a84 in x11_swapchain_destroy () from /nix/store/v6zyaslvymy3pgwqmy7pwn4nhy8a9bkg-mesa-20.3.4-drivers/lib/libvulkan_intel.so
#2  0x0000564737f605ce in ash::vk::extensions::KhrSwapchainFn::destroy_swapchain_khr (self=0x7ffcf091d358, device=..., swapchain=..., p_allocator=0x0)
    at /home/mindtree/.cargo/registry/src/github.com-1ecc6299db9ec823/ash-0.31.0/src/vk/extensions.rs:612
[...]

PR #833, opened 2022-01-14
Fix #826 / Use async-std instead of futures

### LINKED ISSUE #826: "Cannot execute LocalPool executor from within another executor" `panic!` on `master`
A user on matrix mentioned they're running into the following `panic!` on Windows 10:

```
thread 'main' panicked at 'cannot execute LocalPool executor from within another executor: EnterError'.
```

Another user confirmed, and mentioned that the recently released `0.18.1` works, so the issue must be in one of the recent unreleased changes. Seeing as we're getting an executor error, I imagine this must be related to either the new `async` run code or the `wgpu` async code, or maybe the interaction between the two (which I think we always immediately block on anyway). cc @Woyten any ideas on what might be going on? I'll see if I can get some more detailed backtraces, etc.

Woyten:
This PR fixes #826 without reverting the `async` functions.

Since I didn't manage to make the code run using the `futures` library I tried using `async-std` instead which I perceive as a very solid `async` replacement for `std`.

One problem remains: Nannou is not terminating properly. It might have something to do with daemon threads now being non-daemon threads.

### DIFF
--- a/nannou/src/window.rs
+++ b/nannou/src/window.rs
@@ -726,9 +726,10 @@ impl<'app> Builder<'app> {
         self
     }
 
+    #[cfg(not(target_os = "unknown"))]
     /// Builds the window, inserts it into the `App`'s display map and returns the unique ID.
     pub fn build(self) -> Result<Id, BuildError> {
-        futures::executor::block_on(self.build_async())
+        async_std::task::block_on(self.build_async())
     }
 
     pub async fn build_async(self) -> Result<Id, BuildError> {
--- a/nannou_wgpu/src/texture/capturer.rs
+++ b/nannou_wgpu/src/texture/capturer.rs
@@ -27,7 +27,6 @@ pub struct Capturer {
 /// A wrapper around the futures thread pool that counts active futures.
 #[derive(Debug)]
 struct ThreadPool {
-    thread_pool: futures::executor::ThreadPool,
     active_futures: Arc<AtomicU32>,
     workers: u32,
     timeout: Option<Duration>,
@@ -92,7 +91,11 @@ impl ThreadPool {
             active_futures.fetch_sub(1, atomic::Ordering::SeqCst);
         };
 
-        self.thread_pool.spawn_ok(future);
+        #[cfg(not(target_os = "unknown"))]
+        async_std::task::spawn(future);
+        #[cfg(target_os = "unknown")]
+        async_std::task::spawn_local(future);
+
         Ok(())
     }
 
@@ -284,12 +287,7 @@ impl Snapshot {
             .expect("failed to acquire thread handle");
         let thread_pool = guard.get_or_insert_with(|| {
             let workers = self.workers.unwrap_or(num_cpus::get() as u32);
-            let thread_pool = futures::executor::ThreadPoolBuilder::new()
-                .pool_size(workers as usize)
-                .create()
-                .expect("failed to create thread pool");
             let thread_pool = ThreadPool {
-                thread_pool,
                 active_futures: Arc::new(AtomicU32::new(0)),
                 workers,
                 timeout: self.timeout,
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -14,8 +14,8 @@ homepage = "https://github.com/nannou-org/nannou"
 edition = "2018"
 
 [dev-dependencies]
+async-std = "1.10.0"
 audrey = "0.3"
-futures = "0.3"
 hotglsl = { git = "https://github.com/nannou-org/hotglsl", branch = "master" }
 hrtf = "0.2"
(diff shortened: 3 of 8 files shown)


nannou/src/window.rs
```diff
@@ -726,9 +726,10 @@ impl<'app> Builder<'app> {
         self
     }
 
+    #[cfg(not(target_os = "unknown"))]
```
Woyten, line 729: To understand this it's worth checking the implementation of `async_std::task_block_on`. The return type depends on whether your `target_os` is `"unknown"` or not.

nannou/src/window.rs
```diff
@@ -726,9 +726,10 @@ impl<'app> Builder<'app> {
...
     /// Builds the window, inserts it into the `App`'s display map and returns the unique ID.
     pub fn build(self) -> Result<Id, BuildError> {
-        futures::executor::block_on(self.build_async())
+        async_std::task::block_on(self.build_async())
     }
 
     pub async fn build_async(self) -> Result<Id, BuildError> {
```
Woyten, line 735: We could think about putting `#[cfg(target_os = "unknown")]` or `#[cfg(feature = "wasm-experimental")]` here s.t. the `async` we are flexible to change the `async` concept without introducing breaking changes to "normal" users i.e. those that do not use `wasm-experimental`.

nannou_wgpu/src/texture/capturer.rs
```diff
@@ -61,49 +53,10 @@ struct ConverterDataPair {
 pub struct Rgba8AsyncMappedImageBuffer<'buffer>(wgpu::ImageReadMapping<'buffer>);
 
 impl ThreadPool {
-    /// Spawns the given future if a worker is available. Otherwise, blocks and waits for a worker
-    /// to become available before spawning the future.
-    fn spawn_when_worker_available<F>(&self, future: F) -> Result<(), AwaitWorkerTimeout<F>>
```
Woyten, line 66: I have no idea whether this code is important or not. It looks to me as if something is done manually that should be handled by a thread pool executor.

Woyten: Okay... I see this breaks more things...

Woyten:
> One problem remains: Nannou is not terminating properly. It might have something to do with daemon threads now being non-daemon threads.

@mitchmindtree I can no longer reproduce the problem. Maybe someone else could check if the change works for them?

mitchmindtree:
Nice one @Woyten!

I've tested this locally (linux native) and it seems to work nicely, that includes the app and capture examples :+1: 

> async-std

It would be nice if nannou could avoid picking between a particular async runtime (e.g. tokio, async-std, smol), or at least allow for a user to specify their own (while providing an easy default for users who don't care). That said, until we can write async methods for traits, I don't think there's a practical way of achieving this for nannou without some serious refactoring, and I think you're right in that `async-std` is a pretty solid standard to settle on in the meantime.

> One problem remains: Nannou is not terminating properly

Ah I think you might be running into an unrelated issue https://github.com/nannou-org/nannou/issues/728.

@Woyten would you mind sharing which platform you tested this on? I'm with @JoshuaBatty, we might just test with windows and macos before landing.

melMass:
Hello,

Following a chain of issues I ended up here.
I've just tested all wgpu examples (currently trying all others to see), and they now work well on macOS!

@mitchmindtree feel free to ask me stuffs to test on macOS if needed

Thanks

Woyten:
@mitchmindtree Thanks for checking!

> It would be nice if nannou could avoid picking between a particular async runtime

I am not an expert but I think once you use functions like `spawn(_local)` you will always have to pick an async runtime. For example when using `async-std` you implicitly pick the `async_global_executor` runtime. However, since `spawn(_local)` is only used in `capturer.rs` and the used runtime is invisible to the Nannou user I do not know if this really matters much.

> Ah I think you might be running into an unrelated issue #728.

Yes, seems to be it. Phew! :smiley: 

> @Woyten would you mind sharing which platform you tested this on? I'm with @JoshuaBatty, we might just test with windows and macos before landing.

Unfortunately I cannot deliver any new datapoint since I am using Linux as well. But @melMass tested in on macOS. Thank you!

JoshuaBatty:
I just tested on Mac OS and it all seems to be running fine.

Also tested on Windows. Everything compiles and runs fine except the `draw_capture` example seems to be running 4x quicker than the `draw` example. I do have my monitor's refresh rate set to 240Hz just FYI. Running `draw` seems to be updating at normal speed however.

mitchmindtree:
Nice, time to land it!

>  4x quicker than the draw example. I do have my monitor's refresh rate set to 240Hz

It looks like this is just because `draw_capture` uses a fixed time-step interval (`frame.nth() / 60.0`).

PR #863, opened 2022-06-04
Helios laser dac

seem-less:
**Motivation**
To add support for the laser DAC that I have and to gain a deeper understanding of Rust.

**Summary of changes**
Enabling users to be able to use the Helios DAC with the ***nannou_laser*** crate. Information on how to take advantage of these changes is outlined in the [README](https://github.com/seem-less/nannou/blob/bd16e5e1aede228be82aa9bb8de75e5a68dfd948/nannou_laser/README.md?plain=1#L38) and [laser_frame_stream_gui.rs](https://github.com/seem-less/nannou/blob/bd16e5e1aede228be82aa9bb8de75e5a68dfd948/examples/laser/laser_frame_stream_gui.rs#L170).

### DIFF
--- a/nannou_laser/src/dac_manager.rs
+++ b/nannou_laser/src/dac_manager.rs
@@ -1,32 +1,9 @@
-//! Implementation of the general laser point type used throughout the crate.
+//! DAC detection items shared among all supported DACs
+
 use std::io;
-use std::io::ErrorKind;
 use thiserror::Error;
-use helios_dac::{NativeHeliosError, NativeHeliosDac, NativeHeliosDacParams, NativeHeliosDacController};
 
-use crate::dac_manager_helios;
-use crate::{dac_base::DacBase, dac};
 pub type Result<T> = std::result::Result<T, DetectedDacError>;
-
-/// DAC discovery Iterators for all supported DACs available on the system.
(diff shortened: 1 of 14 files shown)


mitchmindtree:
Cheers heaps for this @seem-less, exciting to see helios support :)

Will try get around to reviewing this soon!

nannou_laser/src/dac_manager.rs
```diff
@@ -0,0 +1,163 @@
...
+pub enum DetectDacs {
+    EtherDream {
+        dac_broadcasts: ether_dream::RecvDacBroadcasts,
+    },
+    Helios {
+        previous_dac: helios_dac::NativeHeliosDacParams,
+    },
+}
+
+impl DetectDacs {
```
seem-less, line 18: Looking over this again, it would probably be better if `DetectDacs` and `DetectedDac` impl blocks were refactored as traits allowing for their logic to be placed into their respective files.

seem-less: Attempting to abstract each DAC seemed to add no value and in some instances, dynamic dispatch had to be used with down-casting in order retrieve the required methods needed in the stream routines. I think it's fine to leave these objects as enums rather than refactoring them with traits.

Issue #743, opened 2021-06-01
`app.quit()` causes panics


ErikNatanael:
Using the approach in the draw_capture_hi_res example of waiting for captured frames doesn't work if you want to manually quite the program at some point. `quit` seems to only destroy the windows, but leaves the run loop running which depends on there being a window, as does the `exit` callback in question:
https://github.com/nannou-org/nannou/blob/master/examples/draw/draw_capture_hi_res.rs#L190

Exiting with the ESCAPE key does things in the opposite order:
1. Run the exit callback
2. Return from the run_loop
3. Cleanup?
Reference: https://github.com/nannou-org/nannou/blob/master/nannou/src/app.rs#L1321

Waiting manually by calling the code in the `exit` callback and then calling `app.quit()` sometimes works and sometimes panics.
```
thread 'main' panicked at 'no window for redraw request ID', /home/USERNAME/.cargo/registry/src/github.com-1ecc6299db9ec823/nannou-0.16.0/src/app.rs:761:14
```

It would be useful if the behaviour of app.quit() was consistent with hitting the ESCAPE key e.g. by setting a flag that is checked in the run_loop.

mitchmindtree:
Thanks for opening this @ErikNatanael !

I remember when #712 landed thinking that it was a clever hack to add the `quit` functionality without needing any extra flags or logic, however I think needing access to windows in `exit` is a great case against this approach.

>  quit seems to only destroy the windows, but leaves the run loop running which depends on there being a window, as does the exit callback in question:

Just to clarify, your issue is specifically with the `exit` callback right? I don't think any other event loop callbacks should be called after `app.quit()` is called (though if called in an `update`, it's possible the following `view` is called before `exit`ing - can't quite remember exactly).

I think the proper fix is likely to *not* clear all windows on `app.quit()` and instead set some dedicated flag that can be checked [here](https://github.com/nannou-org/nannou/blob/master/nannou/src/app.rs#L1705) so that windows are still available when `exit` is called and are only cleaned up when `drop` runs afterwards. Does it sound like a solution like this might work for your case?

ErikNatanael:
Thanks for having a look @mitchmintree !

> > quit seems to only destroy the windows, but leaves the run loop running which depends on there being a window, as does the exit callback in question:
> 
[...]

PR #874, opened 2022-07-09
Add a method for fetching the main window when it might not exist

Frojdholm:
I was trying to work with code similar to the `draw_captur_hi_res.rs` example that rendered to a texture, but my code was crashing when I was closing the nannou window using the Windows "X" button. It turns out that it was panicing in the exit function trying to fetch the main window which seems to have closed before the exit function had a chance to run.

I'm not sure if this is the correct solution. Ideally I would change the API of either `App::window_id` or `App::main_window` to return an `Option`. Changing `main_window` would likely not be great since it would add a lot of boilerplate to the more common case of accessing it outside of the exit function. Changing `window_id` would require us to check whether the `focused_window` still existed in `windows`. Both of these would be breaking API changes, which is why I choose to implement new methods.

Also in my particular case it would be enough to check if `window_count` was 0, though that feels a bit brittle in the general case. Additionally, a real workaround would be to check if `App::window_id` exist in `App::windows`. That would work since it looks like `App::focused_window` is set the first time a window is created. In that case I think the docs of `App::window_id` should be updated to reflect that this is something that is guaranteed.

### DIFF
--- a/examples/draw/draw_capture_hi_res.rs
+++ b/examples/draw/draw_capture_hi_res.rs
@@ -187,12 +187,13 @@ fn view(_app: &App, model: &Model, frame: Frame) {
 // Wait for capture to finish.
 fn exit(app: &App, model: Model) {
     println!("Waiting for PNG writing to complete...");
-    let window = app.main_window();
-    let device = window.device();
-    model
-        .texture_capturer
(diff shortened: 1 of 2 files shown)


Frojdholm:
Seem to be a little bit connected to #743

PR #876, opened 2022-07-17
Fix broken example link in readme

nate-sys:
The second GIF in the README was leading to a non-existing page


JoshuaBatty:
Thanks for the PR!

PR #892, opened 2022-10-18
Implement public Error types using thiserror

frankplow:
This PR implements the [`Error`](https://doc.rust-lang.org/std/error/trait.Error.html) trait for [`DrawError`](https://github.com/nannou-org/nannou/blob/0eee95c054ffb08287e0a0c3203f3f0108ddc9e5/nannou/src/draw/renderer/mod.rs#L138) using the `derive` macro provided by [thiserror](https://crates.io/crates/thiserror). This requires adding the [thiserror dependency](https://github.com/frankplow/nannou/blob/561f20d8504774f07a00315477de2ef57ad3585a/nannou/Cargo.toml#L31) to `nannou`. This is in keeping with how `Error`s are implemented elsewhere in the project however - `thiserror` is already used in `nannou_audio`, `nannou_laser` and `nannou_isf`.

### DIFF
--- a/nannou/src/io.rs
+++ b/nannou/src/io.rs
@@ -3,28 +3,24 @@
 
 use serde;
 use serde_json;
-use std::error::Error;
 use std::io::{Read, Write};
 use std::path::Path;
-use std::{fmt, fs, io};
+use std::{fs, io};
+use thiserror::Error;
 use toml;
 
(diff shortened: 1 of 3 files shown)


frankplow:
742b86b extends the thiserror approach to the other public `Error` types from nannou. This reduces boilerplate and increases consistency across the codebase.

Issue #810, opened 2021-12-02
draw_arrow example panics


XBagon:
When moving the mouse over the exact center of an arrow, the example crashes due to a panic.
[crash.log](https://github.com/nannou-org/nannou/files/7640093/crash.log)

mitchmindtree:
Thanks for this, it looks like the crux of the issue is here:

```
RUST_BACKTRACE=1 cargo run --release --example draw_arrow
    Finished release [optimized] target(s) in 0.36s
     Running `target\release\examples\draw_arrow.exe`
thread 'main' panicked at 'not implemented', nannou\src\draw\mesh\builder.rs:91:9
```

which is an `unimplemented!` panic in the `GeometryBuilder::abort_geometry` implementation. We might need to investigate exactly what conditions can cause a lyon geometry builder to abort and handle it accordingly here, whether that's ignoring it or clearing the mesh, etc.

nidnogg:
Commenting out that `unimplemented!` panic, I'm getting a `failed to tessellate path: UnsupportedParamater` on my terminal. Funnily enough though I can't find out where `abort_geometry` is being called. 

Is that possibly called from inside a git submodule or a crate dependency?

PR #894, opened 2022-11-01
Removed panic when mouse is on origin of arrow. Solves #810

Tomostfe:
Solves issue #810 because it does not panic at value 0 or close to 0.


JoshuaBatty:
thanks!

Issue #793, opened 2021-09-18
Make `nannou::app` receive a closure. 


alexfertel:
Currently, `nannou::app` and, in general, the `app::Builder` API expect an argument of type `fn`, but it would be nice to be able to pass a closure instead.

My use case is that I'm writing a CLI wrapper to speed up my workflow, and I need to close over the environment in `model`, `update`, and `view`, but I can't because they have to be of type `fn`.

I think this wouldn't be a breaking change because you can pass a function in a place where a closure is expected, but not the other way around. However, there may be some nuance I am not aware of.

akriegman:
Hi Alex, I had a similar problem to you. What I did is I used the `lazy_static` package so my command line arguments could be in the global scope. I prefer this design pattern personally.

I think your suggestion is probably the right thing to do. I had a [similar problem](https://github.com/nannou-org/nannou/issues/794) with the `draw.polyline().points()` function which takes an `IntoIterator` over `Point2`'s, which means that I have to move ownership of my list of points every time I draw a path, which in my case meant cloning the list every frame. There seems to be a problem in Nannou of functions that don't take exactly the right types and therefore needlessly exclude certain use cases. I'm not sure whether this is a problem with Nannou or a problem with Rust.

alexfertel:
Ooh, I haven't thought of that, I will try it out, thank you!

vvcaw:
Hi, has there been any updates on this issue? I think i stumbled over a similar problem, but please correct me if I'm wrong. I'd like to use some data inside the `model` function inside of an implementation block and came up with this approach.

```Rust
impl Renderer {
    pub fn preview(&mut self) {
        nannou::app(self.model())
            .update(Renderer::update)
            .simple_window(Renderer::view)
            .run();
    }

    fn model(&self) -> fn(app: &nannou::App) -> Animation {
        move |app: &nannou::App| self.animation
    }
}
```

But as `nannou::app` requires a `fn` and closures can't be converted to function pointers if they capture variables I seem to be stuck. Am I missing something or is there no easy fix for this problem?

thinkrapido:
use my fork until then https://github.com/i-think-rapido/nannou/tree/feature/tr/functions-as-closures

vvzen:
+1 - it's really a pity that one has to use global state when closures would have been enough.

[...]

PR #899, opened 2022-11-28
making default functions to closures

thinkrapido:
I like to suggest a pull request to make the default functions like ModelFn, UpdateFn, etc... usable as a closure.

Related: https://github.com/nannou-org/nannou/issues/793

It is a very easy fix and we gain much more flexibility in developing complex programs.

A call would be via an macro
```rust
fn main() {
    nannou::sketch(view!(|app: &App, frame: Frame| {
        //...
    })).run()
}
```
and to respective functions.

Also read, how it is meant: https://medium.com/@romeo-disca/smart-pointers-in-rust-158046006f15

---
<a href="https://gitkraken.dev/review/github/nannou-org/nannou/pull/899">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://gitkraken.dev/images/figures/gitkraken-review-badge-dark.svg">
    Open with GitKraken
  </picture>
</a>

### DIFF
--- a/nannou/src/app.rs
+++ b/nannou/src/app.rs
@@ -31,22 +31,74 @@ use winit;
 use winit::event_loop::ControlFlow;
 
 /// The user function type for initialising their model.
-pub type ModelFn<Model> = fn(&App) -> Model;
+pub type ModelFn<Model> = Arc<dyn Fn(&App) -> Model>;
+#[macro_export]
+macro_rules! model {
+    ($e:expr) => {
+        { std::sync::Arc::new($e) }
+    }
+}
+
 
 /// The user function type for updating their model in accordance with some event.
-pub type EventFn<Model, Event> = fn(&App, &mut Model, Event);
+pub type EventFn<Model, Event> = Arc<dyn Fn(&App, &mut Model, Event)>;
+#[macro_export]
+macro_rules! event {
+    ($e:expr) => {
+        {
+            use std::sync::Arc;
+            Arc::new($e)
+        }
+    }
+}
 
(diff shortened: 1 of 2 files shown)


thinkrapido:
all the examples need to be improved due to this pull request

thinkrapido:
Arc or Box is debatable

Woyten:
Using closures instead of `fn`s would make things easier, indeed. In my personal project, for example, I am using `thread_local!` to pass a pre-computed value to the `fn`. However, there are certain things in your PR I do not understand:
- Why are the types `Arc<dyn Fn(...)>`? Shared `Fn` closures are a pretty rare thing in Rust. You probably want to use `Box<FnMut(...)>`.
- What is the reason to use macros? You could just use regular generics instead e.g. `fn foo(foo_fn: impl FnMut()) { let boxed = Box::new(foo_fn) }`. If you want some kind of name aliasing you could do the following:
```rust
trait FooFn {
    fn foo(&mut self);
}

impl<F: FnMut()> FooFn for F {
    fn foo(&mut self) {
        self()
    }
}

fn consumes_foo(foo_fn: impl FooFn) {
    let boxed: Box<dyn FooFn> = Box::new(foo_fn);
}

fn calls_foo() {
    consumes_foo(|| {})
}
```

abey79:
@thinkrapido Did you make any progresses with this PR? I just ran into the issue of passing state from outside `nannou::app()`, which it would solve.

thinkrapido:
@abey79 , no I didn't. For me the pull request solved to problem.

PR #902, opened 2022-12-19
Fix broken links in examples/README.md

jacksonhenry3:
Fixed links to "simple_ui", "wgpu_triangle", "wgpu_image_sequence", and "wgpu_teapot". For the wgpu examples I also changed the link to direct to the folder, rather than the .rs file, to make it more clear that the shader files are also needed to run the examples.


JoshuaBatty (approved):
Thanks!

Issue #903, opened 2023-01-20
Handle unhandled events introduced in winit 0.27+ (TouchMagnify, IME, etc) in event.rs


mitchmindtree:
Currently these are ignored and return `None`.

hugcis:
Hi, I just opened a PR that updates winit to 0.27 across nannou (#904). 

The TouchMagnify event is not currently part of winit, but I added handling of the IME and Occluded events. If you can check it out, it would be great!

PR #904, opened 2023-01-23
Update to winit 0.27. Handle new window events (IME and Occluded)

hugcis:
## Motivation
This pull request addresses issue #903 and updates the nannou and nannou_egui libraries to be compatible with winit version 0.27+. 

One breaking change in winit 0.27 is the removal of the `WindowAttributes` field on `WindowBuilder.` As a result, additional fields were added to `nannou::window::Builder` to track properties such as `inner_size` and `min_inner_size` before building the window. 

The winit update also introduces two new events: IME and Occluded. 
- The IME event corresponds to a "composition event" (e.g., a series of keyboard events that should be combined in a single produced character, like accented characters) and will only be created if `Window::set_ime_allowed` is called. I ignored the event because the added complexity would outweigh the benefits, but I may be wrong.
- The Occluded event is a macOS-only (for now) event that fires only when the current window is completely hidden from view. It contains a `bool` which I believe indicates if the window is hidden or not (this is not documented in winit). This event can be useful to save computations when the window is not visible.

## Summary of changes: 
- Update to winit 0.27. Update window.rs to deal with a few breaking changes.
- Handle new events IME (ignored) and Occluded (handled).
- Update examples to handle new Occluded events.
- Update changelog.

### DIFF
--- a/examples/nannou_basics/all_functions.rs
+++ b/examples/nannou_basics/all_functions.rs
@@ -33,6 +33,7 @@ fn model(app: &App) -> Model {
         .focused(window_focused)
         .unfocused(window_unfocused)
         .closed(window_closed)
+        .occluded(occluded)
         .build()
         .unwrap();
     Model {}
@@ -79,6 +80,7 @@ fn window_event(_app: &App, _model: &mut Model, event: WindowEvent) {
         Focused => {}
         Unfocused => {}
         Closed => {}
+        Occluded(_val) => {}
     }
 }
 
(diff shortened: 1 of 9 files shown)


mitchmindtree:
Hey thanks so much for the PR @hugcis!

I actually began addressing this update over the weekend as a part of an update to wgpu 0.14 and egui 0.20 :sweat_smile: you can find my WIP branch [here](https://github.com/nannou-org/nannou/compare/master...mitchmindtree:nannou:wgpu-0.14?expand=1).

I plan to resume work on this over the coming weekend, though we're encountering some issues with the way `egui`'s new `egui_wgpu` crate's fragment shader behaves.

hugcis:
Hey @mitchmindtree,

No problem, I am happy to help with nannou! I see you are using an unmerged PR of winit (rust-windowing/winit#2613), which allows skipping some of the workarounds I had to make to update to 0.27. 

It probably makes more sense to update winit with wgpu and egui in one go.

archshift:
Any plans to revisit this?

PR #922, opened 2023-05-30
Try upgrade

anselanza:
This is a messy attempt to get Nannou up to WGPU v0.15 (from 0.11).

After messing around with dependencies (particularly for egui) and adjusting WGSL shader code (the syntax has changed somewhat in the last few years) I have got to the point where Nannou compiles at least for a simple example like `draw` 

But I think I'm missing some understanding of the details of how Nannou translates WGPU textures, etc. so I get the following error message:

```
Caused by:
    In a RenderPass
      note: encoder = `nannou_raw_frame`
    In a set_pipeline command
      note: render pipeline = `nannou render pipeline`
    Render pipeline targets are incompatible with render pass
    Incompatible color attachment: the renderpass expected [Some(Rgba8Unorm)] but was given []

', /Users/stephen/.cargo/registry/src/github.com-1ecc6299db9ec823/wgpu-0.15.1/src/backend/direct.rs:3024:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'main' panicked at 'wgpu error: Validation Error

Caused by:
    In a CommandEncoder
    command encoder is invalid
```
Any help from more experienced Rust+Nannou devs would be much appreciated.

### DIFF
--- a/examples/laser/laser_frame_stream_gui.rs
+++ b/examples/laser/laser_frame_stream_gui.rs
@@ -534,7 +534,7 @@ fn fonts() -> egui::FontDefinitions {
             (egui::FontFamily::Monospace, 14.0),
         ),
     ];
-    fonts.family_and_size.extend(entries.iter().cloned());
+    // fonts.family_and_size.extend(entries.iter().cloned());
     fonts
 }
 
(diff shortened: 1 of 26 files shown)


anselanza:
Oh yes, winit version had to change, as well: 0.27 instead of 0.26

PR #861, opened 2022-05-29
Update to wgpu 0.12, egui 0.18

mitchmindtree:
Finally getting around to updating these!

The `wgpu` update is quite trivial and looks like the new validation errors might make it easier to catch bugs in the WIP `nannou_isf` crate.

The `egui` update however needs a more significant refactor. It looks like the `egui_winit_platform` and `egui_wgpu_backend` crates have recently been merged into the main `egui` repo as `egui-winit` and `egui-wgpu` respectively. Awesome to see these being maintained upstream, but both have transformed quite a bit and require some rethinking of `nannou_egui` and `nannou_egui_demo_app`.

## TODO

- [x] **Update `nannou_wgpu` to `wgpu` `0.12`.**
- [x] **Update `nannou` to `wgpu` `0.12`.**
- [x] **Update `nannou_isf` to `wgpu` `0.12`.**
- [ ] **Update `nannou_egui` to `egui` `0.18`.** Mostly done, though need to refactor eframe stuff for demo app.
- [ ] **Update `nannou_egui_demo_app`.** Changed significantly, might need to copy over quite a bit of boilerplate from `egui_demo_app`.
- [ ] **Update `wgpu` examples to `0.12`.**

PR #940, opened 2023-10-10
Library upgrades and updates

tychedelia:
Mostly working, but need to double check a few things around egui. And make sure all the shaders are still working with wgsl changes.

- Switch from `async-std` to `tokio`. 
- Switch from `egui_wgpu_backend` to the in-tree `egui-wgpu`.
- Upgrade to `egui` version `0.23`.
- Upgrade to `wgpu` version `0.17`.
- Upgrade to `winit` version `0.28`.

TODO:
- [x] Ensure all targets (i.e. WASM) still working.
- [x] Figure out whether we can use `eframe` as a replacement for `epi` in the new egui version. (answer: no)
- [x] Resolve issue from https://github.com/nannou-org/nannou/pull/940#discussion_r1365303258 in the `draw_capture_hi_res` example. 
- [ ] Figure out resize issue on Ubuntu.
- [ ] Establish egui pattern to replace epi.
- [ ] Egui demo colors.


infinitylunacreative:
I can confirm that egui seems to be working correctly. I will let you know if anything comes up.

infinitylunacreative:
Wasm builds now break. 

I think this is due tokio+mio - seems to breaks on WASM

`  
 Compiling mio v0.8.8
error[E0432]: unresolved import `crate::sys::IoSourceState`
`

infinitylunacreative:
@tychedelia I attempted to test WASM compiling by pulling out tokio. But got suck (new to rust) - I have left a PR on your repo it might help it might not :)

infinitylunacreative:
> > @tychedelia I attempted to test WASM compiling by pulling out tokio. But got suck (new to rust) - I have left a PR on your repo it might help it might not :)
> 
> Thanks! I totally blanked on testing WASM. I'll give your change a shot today or otherwise make sure things are working with the WASM builds.

No problem, Im newish to rust (:sweat_smile: but learning fast), I got stuck. The PR to yours is as far as I could get.

tychedelia:
@infinitylunacreative just pushed some fixes that should allow tokio to run on wasm targets. will you give it a shot and let me know if that works (on the head of my fork)?

tychedelia:
Looks like @mitchmindtree ran into some of the same questions re: `eframe` in this previous upgrade attempt here https://github.com/nannou-org/nannou/pull/861. It does appear that the upstream might be willing to accept changes to `eframe` to make it work better with our use case. It would be wonderful if there was a more fully featured example of a nannou app that uses the previous `epi` functionality, but I couldn't find one.

infinitylunacreative:
> @infinitylunacreative just pushed some fixes that should allow tokio to run on wasm targets. will you give it a shot and let me know if that works (on the head of my fork)?

Thanks, I will take a look today.

infinitylunacreative:
@tychedelia I can confirm the WASAM build is working for me 🥳 - Thank you!

infinitylunacreative:
@tychedelia I decided to do some more testing. I tried it on linux. WASAM works great. 

However the "normal" builds throw this error.

error[E0432]: unresolved import `winit::platform::unix`
   --> /home/infinity/work/nannou/nannou/src/window.rs:783:34
    |
783 |             use winit::platform::unix::WindowBuilderExtUnix;
    |                                  ^^^^ could not find `unix` in `platform`

infinitylunacreative:
@tychedelia I decided to do some more testing. I tried it on linux. WASAM works great. 

However the "normal" builds throw this error.

```
error[E0432]: unresolved import `winit::platform::unix`
   --> /home/infinity/work/nannou/nannou/src/window.rs:783:34
    |
783 |             use winit::platform::unix::WindowBuilderExtUnix;
    |                                  ^^^^ could not find `unix` in `platform`
```

```
error[E0599]: no method named `with_class` found for struct `WindowBuilder` in the current scope
   --> /home/infinity/work/nannou/nannou/src/window.rs:784:29
    |
784 |             window = window.with_class("nannou".to_string(), "nannou".to_string());
    |                             ^^^^^^^^^^ method not found in `WindowBuilder`
```

tychedelia:
> @tychedelia I decided to do some more testing. I tried it on linux. WASAM works great.
> 
> However the "normal" builds throw this error.
> 
> ```
> error[E0432]: unresolved import `winit::platform::unix`
>    --> /home/infinity/work/nannou/nannou/src/window.rs:783:34
>     |
> 783 |             use winit::platform::unix::WindowBuilderExtUnix;
>     |                                  ^^^^ could not find `unix` in `platform`
> ```
> 
> ```
> error[E0599]: no method named `with_class` found for struct `WindowBuilder` in the current scope
>    --> /home/infinity/work/nannou/nannou/src/window.rs:784:29
>     |
> 784 |             window = window.with_class("nannou".to_string(), "nannou".to_string());
>     |                             ^^^^^^^^^^ method not found in `WindowBuilder`
> ```

Looks like there was a breaking change in `winit`: https://github.com/rust-windowing/winit/blob/master/CHANGELOG.md?plain=1#L255.

Not sure the best approach here. I think I'm going to enable xorg by default and make wayland an additional feature flag.

tychedelia:
Should be fixed now, but I haven't tested wayland.

infinitylunacreative:
Can con firm it is fixed. - Thank you.

tychedelia:
> Can con firm it is fixed. - Thank you.

Thank you for all the testing!!

infinitylunacreative:
> > Can con firm it is fixed. - Thank you.
> 
> Thank you for all the testing!!

You are welcome. But I also really want this update and I am really enjoying Rust/nannou over Processing and OpenFrameworks. 

One thing I just noticed is there is a "space" around the whole window on the Linux build windows vs WASAM.  In these screen shots I am using the window rect. On a 800x800 sketch. I dont know if this is a setting or a bug? - I think a bug :shrug: 

Haven't checked OSX yet. I will post when I do.

Screenshot from 2023-10-14 09-31-18

Screenshot from 2023-10-14 09-38-26

tychedelia:
> One thing I just noticed is there is a dimension "space/padding" around the whole window on the Linux build windows vs WASAM. But the canvas extends past as you can see in my squares on the right of the window screenshot.

Yeah, this seems like a bug. Let's see if we can fix this... Would you mind sharing your example sketch so I can try to reproduce? The few examples I've tried don't seem to have this padding on a fresh copy of Ubuntu 22.04.2 ARM64.

infinitylunacreative:
Here is a repo with a base sketch in in: 

https://github.com/infinitylunacreative/nannou_update_test

In linux you can see the cords are at -410, 410, on an 800,800. 

image

infinitylunacreative:
I am getting the following when using reshaper

#### Stack trace 
```
/home/infinity/work/creative_code/nn_001/target/debug/nn_001_compat
thread 'main' panicked at 'wgpu error: Validation Error

Caused by:
    In Device::create_shader_module
      note: label = `shaders/fs_msaa4.wgsl`
    
Shader 'shaders/fs_msaa4.wgsl' parsing error: the type of `tex_size` is expected to be `vec2<i32>`, but got `vec2<u32>`
   ┌─ wgsl:15:9
   │
15 │     let tex_size: vec2<i32> = textureDimensions(tex);
   │         ^^^^^^^^ definition of `tex_size`


    the type of `tex_size` is expected to be `vec2<i32>`, but got `vec2<u32>`

', /home/infinity/.cargo/registry/src/index.crates.io-6f17d22bba15001f/wgpu-0.17.1/src/backend/direct.rs:3056:5
stack backtrace:
   0: rust_begin_unwind
             at /rustc/eb26296b556cef10fb713a38f3d16b9886080f26/library/std/src/panicking.rs:593:5
   1: core::panicking::panic_fmt
             at /rustc/eb26296b556cef10fb713a38f3d16b9886080f26/library/core/src/panicking.rs:67:14
   2: wgpu::backend::direct::default_error_handler
             at /home/infinity/.cargo/registry/src/index.crates.io-6f17d22bba15001f/wgpu-0.17.1/src/backend/direct.rs:3056:5
   3: core::ops::function::Fn::call
             at /rustc/eb26296b556cef10fb713a38f3d16b9886080f26/library/core/src/ops/function.rs:79:5
   4: <alloc::boxed::Box<F,A> as core::ops::function::Fn<Args>>::call
             at /rustc/eb26296b556cef10fb713a38f3d16b9886080f26/library/alloc/src/boxed.rs:1999:9
   5: wgpu::backend::direct::ErrorSinkRaw::handle_error
             at /home/infinity/.cargo/registry/src/index.crates.io-6f17d22bba15001f/wgpu-0.17.1/src/backend/direct.rs:3042:17
   6: wgpu::backend::direct::Context::handle_error
             at /home/infinity/.cargo/registry/src/index.crates.io-6f17d22bba15001f/wgpu-0.17.1/src/backend/direct.rs:324:9
   7: <wgpu::backend::direct::Context as wgpu::context::Context>::device_create_shader_module
             at /home/infinity/.cargo/registry/src/index.crates.io-6f17d22bba15001f/wgpu-0.17.1/src/backend/direct.rs:931:13
   8: <T as wgpu::context::DynContext>::device_create_shader_module
             at /home/infinity/.cargo/registry/src/index.crates.io-6f17d22bba15001f/wgpu-0.17.1/src/context.rs:2241:37
   9: wgpu::Device::create_shader_module
             at /home/infinity/.cargo/registry/src/index.crates.io-6f17d22bba15001f/wgpu-0.17.1/src/lib.rs:2251:26
  10: nannou_wgpu::texture::reshaper::Reshaper::new
             at /home/infinity/work/nannou/nannou_wgpu/src/texture/reshaper/mod.rs:54:22
  11: nn_001_compat::sketch_model::Model::build_woker
             at ./src/sketch_model.rs:116:32

```
This is from code the Hres in examples/drawing. 
```
      let texture_reshaper = wgpu::TextureReshaper::new(
            device,
            &texture_view,
            sample_count,
            texture_sample_type,
            sample_count,
            dst_format,
        );
 ```
 
 The examples use i32, but the updates seem to use u32. I looked at the shaders to see if I could fix but not to sure which side of the fence to even attempt.

**Comment by RobWalt** [comment]
Real cool iniziative. I'd like to join you two to build up some more momentum since the project seems to be quiet inactive otherwise. I'll give the branch some tests and a review as well as soon as I'm on the PC.

The last panic looks like a type error in the shader and should be rather easy to fix.

**Comment by infinitylunacreative** [comment]
@RobWalt that would be great.  It is all @tychedelia. Im just here breaking things 😬 🤣 .

I am new to shaders as a whole and don't really know how to fix, but yes please get involved. I am guessing fixes you put in should come as a PR to this working repo. This way hopefully we can get it all squared away.

**Comment by tychedelia** [comment]
@RobWalt  Going to look at some of these today, but feel free to PR into my fork (aka this PR)!

File: nannou_wgpu/src/texture/image.rs [hunk 1]
```diff
@@ -311,7 +312,14 @@ impl wgpu::RowPaddedBuffer {
...
         let slice = self.buffer.slice(..);
-        slice.map_async(wgpu::MapMode::Read).await?;
+        let (tx, rx) = futures::channel::oneshot::channel();
+
+        slice.map_async(wgpu::MapMode::Read, |res| {
+            tx.send(res).expect("Failed to send map_async result");
+        });
+
+        rx.await.expect("Failed to receive map_async result")?;
+
```
RobWalt, line 322: I'm currently working on fixing everything in this comment here https://github.com/nannou-org/nannou/pull/940#issuecomment-1762594523

I was able to fix two things, the commits can be found here

- https://github.com/RobWalt/nannou/commit/648668efdb8b6916c5725121712fa22c26ef44d2
- https://github.com/RobWalt/nannou/commit/9a61ce7943147f08858c1443737cb44bee07d7dc

Now wgpu is complaining a lot about these lines here. The docs of `map_async` state

> Map the buffer. Buffer is ready to map once the callback is called.
>
>For the callback to complete, either `queue.submit(..)`, `instance.poll_all(..)`, or `device.poll(..)`
must be called elsewhere in the runtime, possibly integrated into an event loop or run on a separate thread.
>
>The callback will be called on the thread that first calls the above functions after the gpu work
has completed. There are no restrictions on the code you can run in the callback, however on native the
call to the function will not complete until the callback returns, so prefer keeping callbacks short
and used to set flags, send messages, etc.

and I think we don't use any of the 3 mentioned methods here. Not sure how to proceed at the moment

tychedelia:
> Here is a repo with a base sketch in in:
> 
> https://github.com/infinitylunacreative/nannou_update_test
> 
> Hopefully In linux you can see the cords are at -410, 410, on an 800,800.
> 
> I'm running Ubuntu 22.04.3 64bit
> 
> image

Update:

The first resize event sent by winit incorrectly includes 20px, which seems to be the case regardless of user specified setting. This feels like a potential bug in winit itself, maybe https://github.com/rust-windowing/winit/issues/2094. Next steps are likely to try upgrading to the `29.1-beta` release and see if it's been fixed there. Otherwise I'll file a ticket on winit and see if they have any advice.

nannou_wgpu/src/texture/image.rs
tychedelia: Hmm, okay, good catch. This was a breaking change on the upgrade and may have just been an oversight by me. Previously they returned a future from this `map_async` method but claimed the overhead of polling was too high, hence the callback based version that only fires once. I'm not super familiar with this portion of the render pipeline, but might poke around later tonight to see.

RobWalt: As far as I'm understanding this, we really have to call one of the three methods

> `queue.submit(..)`, `instance.poll_all(..)`, or `device.poll(..)`

to make that work before awaiting the rx (since the rx is one shot). This is an issue because with the current design it's at best very unclear if this is called at all and at worst isn't called yet. For me this seems like it would need a bigger redesign 😬

RobWalt (commented):
@tychedelia What's the motivation behind the switch to `tokio`? I get that `async-std` isn't the latest async runtime. In the context of graphics programming with wgpu I just saw a lot of use of `pollster` instead and maybe that's useful here aswell since we're always just blocking anyways?

Imo it would be better if we split this PR into smaller ones which have a smaller scope. So the async refactor could be its own PR and so on.

anselanza:
I can't offer much help here, but just want to commend the effort to bring Nannou up to date regarding WGPU and Egui. I really like this framework - and have used it in at least one project in production - but I was getting concerned about the apparent lack of updates recently.

tychedelia:
Since I opened this, `winit` has released an `0.29` version which includes their [event loop 3.0 rework[(https://github.com/rust-windowing/winit/releases/tag/v0.29.2) that has a *substantial* number of changes. I've attempted an upgrade, and while there's a lot of little fiddly changes that aren't too bad, there are some substantial changes that are somewhat difficult to integrate.

1. `Key` has been split into an an enum representing differences between character keys and "named" keys. For example, most of the generative design examples require changing their screenshot trigger from:

```rust
fn key_pressed(app: &App, _model: &mut Model, key: Key) {
    if key == Key::S {
        app.main_window()
            .capture_frame(app.exe_name().unwrap() + ".png");
    }
}
```

to 

```rust
fn key_released(app: &App, _model: &mut Model, key: Key) {
    if let Key::Character(key) = key {
        if "s" == key {
            app.main_window()
                .capture_frame(app.exe_name().unwrap() + ".png");
        }
    }
}
```

This isn't a hugeeee deal, but definitely is less ergonomic for our users.

2. LOTS of changes to events emitted by `winit`. The biggest is probably moving `RedrawRequested` to `WindowEvent`. These are probably not too bad to refactor but there's *lots* of changes here and so it's a bit hard to analyze what events might be relevant to users.

It would be *really* nice to be able to land a `29.x` upgrade in order to get nannou up to a stable state with regards to one of its most important dependencies, but this potentially will incur a *lot* of user visible breakage for what's ultimately pretty minimal gain for the user.

tychedelia:
https://github.com/nannou-org/nannou/pull/940/commits/36a724e9783656bf5a197f4b2ece8160f053f017 fixes the issue with the egui demo app, but isn't the right pattern for our users. likely more to do here, it's rendering correctly now though.

nannou_wgpu/src/texture/image.rs
tychedelia: We flush the queue when the frame is dropped. There were some additional issues with the shaders for this, but the example is running for me now just fine.

nannou_egui_demo_app/src/main.rs
```diff
@@ -20,10 +22,11 @@ fn model(app: &App) -> Model {
         .unwrap();
     let window = app.window(w_id).unwrap();
     let mut egui = Egui::from_window(&window);
-    let mut egui_demo_app = egui_demo_lib::WrapApp::default();
-    let proxy = app.create_proxy();
-    egui.do_frame_with_epi_frame(proxy, |ctx, epi_frame| {
-        egui_demo_app.setup(&ctx, epi_frame, None);
+    let mut egui_demo_app = egui_demo_lib::DemoWindows::default();
```
tychedelia, line 25: Not sure the best way to clean this up. Maybe make the `Egui` constructors accept an `App` instance so we can wire that up internally?

nannou_egui_demo_app/src/main.rs
```diff
@@ -42,12 +45,12 @@ fn update(app: &App, model: &mut Model, update: Update) {
         ..
     } = *model;
     egui.set_elapsed_time(update.since_start);
-    let proxy = app.create_proxy();
-    egui.do_frame_with_epi_frame(proxy, |ctx, frame| {
-        egui_demo_app.update(&ctx, frame);
-    });
+    let ctx = egui.begin_frame();
```
tychedelia, line 48: These should probably just be moved into a `do_with_frame` method on `egui` to keep it simple.

mitchmindtree (commented):
Wicked work @tychedelia, I really appreciate you diving into this!

I've opened up a PR with some small tweaks and fixes if you're interested [here](https://github.com/tychedelia/nannou/pull/2) - once those are merged I'd be more than happy to see this land :)

---

As a small side-note, it seems like this PR runs into the same issues that I ran into in #861 w.r.t. colours not quite matching up. From memory, this has something to do with egui changing something related to the way they do gamma correction in their fragment shader and some assumptions they make around web targets vs desktop targets. I think nannou uniquely runs into this as we don't target the window surface texture directly like most egui apps do - instead we draw to an intermediary frame with a linear colour format.

All that said, I think we might be best to:

1. Land this as is (with aforementioned tweaks PR)
2. Open an issue for tracking the egui colour problem.
3. Aim to solve the problem through our bevy refactor when switching from `nannou_egui` to the `bevy_egui` crate.

This still might require some thought around texture colour formats etc, but I think we're better off investing that effort as a part of the bevy-refactor, rather than delaying it longer? Open to other suggestions!

nannou_egui/src/lib.rs
```diff
@@ -435,19 +410,19 @@ impl<'a> Drop for FrameCtx<'a> {
...
-        }
-    }
-}
+// impl epi::RepaintSignal for RepaintSignal {
+//     fn request_repaint(&self) {
+//         if let Ok(guard) = self.0.lock() {
+//             guard.wakeup().ok();
+//         }
+//     }
+// }
```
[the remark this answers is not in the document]
tychedelia: good catch, thanks!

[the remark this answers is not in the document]
mitchmindtree: Oh woops was going to address this in the other PR but I forgot :sweat_smile:

tychedelia:
@mitchmindtree  yeah, seeing the colors are super noticeable here in the color picker example. 

i'm good with this plan! we can fast follow anything that might come up from further user testing.

image

mitchmindtree:
Okydokey let's land this!

I'll follow-up with a PR bumping versions so we can publish to crates.io :+1:

PR #943, opened 2023-11-19
Refac unnecessary `ref mut` pattern matching

marcospb19:
As an attempt to make the code more idiomatic.


marcospb19:
Knowing "idiomatic Rust" is a subjective concept, feel free to close this if you don't see benefit in it.

marcospb19:
I didn't include changes like this:

```rust
// From
if let Some(ref mut dac) = maybe_dac {
// To
if let Some(dac) = &mut maybe_dac {
```

That replace `ref mut` by `&mut`, I prefer the latter, but I guess in that case it doesn't matter A or B.

PR #950, opened 2024-01-17
Add a Nix flake along with a package derivation (default.nix) and dev shell (shell.nix). Update hotglsl.

mitchmindtree:
These can be ignored by most users, but will be useful for Nix folks and may help to simplify our CI a little.

The default.nix can also be used as a CI-checked reference of what system dependencies and env vars are required on Linux and macOS.

Tested successfully on `x86_64-linux` and `aarch64-darwin`.

### TODO

- [x] Rebase onto `bevy-refactor` branch.
- [x] Simplify CI dependency installation with Nix.
- [x] Add section to `Contributing.md` about Nix.

---

The `hotglsl` crate was updated to 0.2 in order to switch from the `glsl-to-spirv` crate to the pure-Rust `naga` removing the need for some system deps like libclang.

### DIFF
--- a/.github/workflows/nannou.yml
+++ b/.github/workflows/nannou.yml
@@ -8,15 +8,6 @@ jobs:
     - uses: DeterminateSystems/nix-installer-action@main
     - run: nix fmt -- --check ./
 
-  cargo-audit:
-    runs-on: ubuntu-latest
-    continue-on-error: true
-    steps:
-      - uses: actions/checkout@v3
-      - uses: rustsec/audit-check@v1.4.1
-        with:
-          token: ${{ secrets.GITHUB_TOKEN }}
-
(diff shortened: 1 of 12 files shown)


tychedelia:
@mitchmindtree Just had to do this for bevy which was kinda annoying https://github.com/nannou-org/nannou/pull/952/commits/e5fcbe95f37ca9171b8826dd38e416c29d33d0ee. The simplest fix would be a [composite action](https://docs.github.com/en/actions/creating-actions/creating-a-composite-action) but I could be convinced to run nix in ci. ^^

JoshuaBatty (approved):

mitchmindtree:
OK I've had a go at simplifying the CI a little and added a "Nix Shell" contributing chapter!

Issue #945, opened 2023-12-24
Memory leak when running examples


olafurkarl:
Good day, when I run any of the examples I get a massive memory leak. I started noticing it due to my system freezing up while diving into them for the first time.

To reproduce

Check out project
* Run example, either from command line following instructions in README.md (`cargo run --release --example <example_name>`), or through rust-analzyer
* Run top to monitor memory usage
* Memory will steadily creep up at a steady pace until all system resources are exhausted.
Running any example reproduces, and running with an empty draw function does as well. I also tried using `::app` rather than `::sketch` with the same result.

rustc version: rustc 1.74.1 (a28077b28 2023-12-04)

PR #960, opened 2024-01-26
#945 Bevy render PoC

tychedelia:
TODO:
- [ ] Support user uploaded textures using the Bevy asset system.
- [ ] MSAA support.
- [ ] Confirm we can use bevy's dpeth texture.
- [ ] Figure out if text still works.
- [ ] Re-implement draw command analog to support switching textures, rendering multiple mesh, scissor, etc.
- [ ] Move `ViewMesh` to render world in prep for draw work.

### DIFF
--- a/bevy_nannou/src/main.rs
+++ b/bevy_nannou/src/main.rs
@@ -1,0 +1,7 @@
+use bevy_nannou::NannouPlugin;
+
+pub fn main() {
+    let mut app = bevy::app::App::new();
+    app.add_plugins((bevy::DefaultPlugins, NannouPlugin));
+    app.run();
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -585,7 +585,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3f48b9bbe4ec605e4910b5cd1e1a0acbfbe0b80af5f3bcc4489a9fdd1e80058c"
 dependencies = [
  "bevy_macro_utils",
- "proc-macro2 1.0.76",
+ "proc-macro2 1.0.78",
  "quote 1.0.35",
  "syn 2.0.48",
 ]
@@ -699,7 +699,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a8843aa489f159f25cdcd9fee75cd7d221a7098a71eaa72cb2d6b40ac4e3f1ba"
 dependencies = [
  "bevy_macro_utils",
- "proc-macro2 1.0.76",
+ "proc-macro2 1.0.78",
  "quote 1.0.35",
  "syn 2.0.48",
 ]
@@ -792,7 +792,7 @@ dependencies = [
  "bevy_log",
  "bevy_reflect",
  "bevy_utils",
- "smallvec 1.12.0",
+ "smallvec 1.13.1",
 ]
 
 [[package]]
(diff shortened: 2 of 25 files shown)


bevy_nannou/src/main.rs
tychedelia, line 1: this isn't necessary long term but is a convenient way to run things for now

tychedelia:
This diff is pretty hard to review, but I wanted to give @mitchmindtree and @JoshuaBatty a chance to review or comment prior to merging this into our `bevy-refactor` branch in preparation for some initial `Draw` work. There's some outstanding issues here that will be easier to fix with more robust examples running. I've tried to mark things with `TODO` but this should still be understood as a first pass.

The main file I would look at is `pipeline.rs`, which outlines the basic flow. There's some stuff currently *missing* there, which is a consequence of lacking the draw api -- for example, right now, we don't even create pipelines (removed in https://github.com/nannou-org/nannou/pull/960/commits/a37611edec2075bb7e249e46d38382684a564114). Happy to answer more questions here, and going to add a few notes to #945 in terms of additional learnings.

bevy_nannou/src/main.rs
```diff
@@ -0,0 +1,82 @@
...
+    commands.spawn(PbrBundle {
+        mesh: meshes.add(Mesh::from(Torus {
+            radius: 100.0,
+            ..default()
+        })),
+        transform: Transform::from_xyz(-200.0, -200.0, 0.0),
+        ..default()
+    });
+
+    commands.spawn(
```
tychedelia, line 32: we should probably spawn the camera for the user since we need to manage something like the orthographic settings and the clear color.

bevy_nannou/src/main.rs
```diff
@@ -0,0 +1,82 @@
...
+#[derive(Resource, Deref, DerefMut)]
+struct MyTexture(Handle<Image>);
+
+fn startup(mut commands: Commands, assets: Res<AssetServer>, mut meshes: ResMut<Assets<Mesh>>) {
+    commands.insert_resource(AmbientLight {
+        color: Color::WHITE,
+        brightness: 1.0,
+    });
+    commands.spawn(PbrBundle {
+        mesh: meshes.add(Mesh::from(Torus {
```
tychedelia, line 24: this was just to test that we draw correctly over bevy meshes (we do)

JoshuaBatty:
Awesome work @tychedelia ! so cool to see this and encouraged by how much of stuff we can offload to bevy so far. I'm happy with the approach taken here, i'm sure there will be some more changes as you go along so don't want to hold this PR and the flow up.

tychedelia:
(Note for self) Right now we're using a `ViewNode` to drive the render, but I think that we might be able to more idiomatically render our mesh using Bevy's `PhaseItem`/`RenderCommand` pattern. See the `bevy_gizmos` crate for an example of what this might look like.

tychedelia:
Waiting on this https://github.com/NixOS/nixpkgs/pull/289940.

JoshuaBatty (approved):

.gitignore
```diff
@@ -7,3 +7,4 @@ target/
 **/*.rs.bk
 .DS_Store
 .idea/
+result/
```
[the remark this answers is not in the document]
mitchmindtree: I think you've done the right thing here! I'm just in the habit of clearing `result` I think and forgot to add it :sweat_smile:

PR #964, opened 2024-03-01
962 use mid level render apis for render

tychedelia:
# Mid-level Render APIs

This pull request moves us from using a roughly 1:1 conversion of the rendering technique on `master`, which uses a custom `ViewNode` that runs on Bevy's 3d render graph, to using a variety of "mid-level" render APIs that bring us closer in line with Bevy's idioms for render code.

## High level overview

Rather than executing a series of render commands, we queue a number of render items into the `Transparent3d` phase of Bevy's existing render graph. These items are then executed with a "draw function", which specifies a series of render operations to perform on each item, setting the pipeline, bind groups, and finally issuing a draw call.

This initially poses a problem, because our existing rendering code executes a variety of draw calls against a single mesh, and so isn't easily broken into "items." Unlike other rendering code Bevy handles, we don't have the benefit of assuming that logical rendering entities exist for multiple frames. In the worst case scenario, each frame is maximally different in every dimension from the one that came before it. This makes patterns in Bevy designed around rendering stable assets a bit more difficult to fit into.

The solution in this PR is the following:
- Keep our existing mesh, which has been renamed from `ViewMesh` to `DrawMesh` and converted to a Bevy asset. We still associate draw instances with a view, but instead let Bevy manage the lifetime of the asset, and interact with a handle to the mesh. This better identifies our mesh as having a relationship to certain GPU resources, which are created in the mesh lifecycle, rather than explicitly in the render code.
- When the mesh is filled, spawn a number of `DrawMeshItem`s that contain metadata that are used to drive our new draw function, like what texture should be used for rendering this portion of the mesh, etc. These are created on every frame and extracted into the render world where they are added to the render phase.
- Moved vertex mode information into another uniform, which was necessary to help decouple the new asset pattern from our previous imperative rendering logic.

## Learning examples in Bevy

These patterns are used all over the renderer, but it's worth calling out a few examples:

- [Gizmos](https://github.com/bevyengine/bevy/blob/main/crates/bevy_gizmos/src/pipeline_3d.rs), which are an immediate style drawing API somewhat similar conceptually to Nannou.
- [The instancing example](https://github.com/bevyengine/bevy/blob/main/examples/shader/shader_instancing.rs), which is pretty verbose right now but potentially relevant to our users.
- [The manual 2d mesh example](https://github.com/bevyengine/bevy/blob/main/examples/2d/mesh2d_manual.rs), which is probably the most straightforward example for understanding some of these patterns.

## How to review

Unfortunately, the diff here got pretty messy, I hope it's still somewhat comprehensible, but I can move things around into new files or otherwise if it would make it clearer or easier to review.

- Please bikeshed names. In particular, things like `DrawDrawMesh` are too overloaded. I was considering keeping `Draw` in terms of our high level API, but calling things `Drawing` in the render code to distinguish it from Bevy's draw/mesh abstractions.
- In `bevy_nannou_render/src/lib.rs`, notice how much code was able to be removed from `update_draw_mesh` (previously `prepare_view_mesh`. Also the `RenderAsset` mechanism for `DrawMesh`/`GpuDrawMesh`.
- `bevy_nannou_render/src/pipeline.rs` should be more straightforward.

## Conclusion and next steps

Some of these patterns are more verbose and a bit abstract, but I think are worthwhile in bringing us closer to Bevy. There's also definitely room for improvement here, including just more general cleanup and tidiness. In particular, it's still worth investigating whether we can make more use of Bevy's mesh infrastructure, rather than re-inventing the wheel here. This might not be possible due to our immediate mode needs, but is worth ruling out.

### DIFF
--- a/bevy_nannou_render/src/pipeline.rs
+++ b/bevy_nannou_render/src/pipeline.rs
@@ -1,42 +1,28 @@
 use std::hash::Hash;
 
-use bevy::core_pipeline::core_3d::{Transparent3d, CORE_3D_DEPTH_FORMAT};
-use bevy::ecs::query::ROQueryItem;
-use bevy::ecs::system::lifetimeless::{Read, SRes};
-use bevy::ecs::system::SystemParamItem;
-use bevy::pbr::{MeshPipeline, MeshPipelineKey, SetMeshViewBindGroup};
+use bevy::core_pipeline::core_3d::CORE_3D_DEPTH_FORMAT;
+use bevy::pbr::{MeshPipeline, MeshPipelineKey};
 use bevy::prelude::*;
-use bevy::render::extract_component::DynamicUniformIndex;
-use bevy::render::render_asset::RenderAssets;
-use bevy::render::render_phase::{
-    DrawFunctions, PhaseItem, RenderCommand, RenderCommandResult, RenderPhase, SetItemPipeline,
-    TrackedRenderPass,
-};
+use bevy::render::render_phase::RenderCommand;
 use bevy::render::render_resource as wgpu;
-use bevy::render::render_resource::{
-    PipelineCache, RenderPipelineDescriptor, SpecializedRenderPipeline, SpecializedRenderPipelines,
-};
+use bevy::render::render_resource::{RenderPipelineDescriptor, SpecializedRenderPipeline};
 use bevy::render::renderer::RenderDevice;
-use bevy::render::texture::BevyDefault;
-use bevy::render::view::{ExtractedView, ViewTarget};
+use bevy::render::view::ViewTarget;
 
 use bevy_nannou_draw::draw::mesh;
 use bevy_nannou_draw::draw::mesh::vertex::Point;
 
-use crate::{
-    DefaultTextureHandle, DrawMesh, DrawMeshHandle, DrawMeshItem, DrawMeshUniform,
-    DrawMeshUniformBindGroup, TextureBindGroupCache, NANNOU_SHADER_HANDLE,
-};
+use crate::{DrawMeshUniform, NANNOU_SHADER_HANDLE};
 
 #[derive(Resource)]
 pub struct NannouPipeline {
-    mesh_pipeline: MeshPipeline,
-    glyph_cache_texture: wgpu::Texture,
-    text_bind_group_layout: wgpu::BindGroupLayout,
-    text_bind_group: wgpu::BindGroup,
-    pub(crate) texture_bind_group_layout: wgpu::BindGroupLayout,
-    texture_bind_group: wgpu::BindGroup,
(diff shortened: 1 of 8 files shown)


tychedelia:
Realizing that one issue with this design is that we are setting the same vertex buffer over and over for each phase item. Wondering if we should either:
- Set the correct ranges of vertex data for each item.
- Rethink this approach and spawn an arbitrary set of render assets matching our phase items per frame.

tychedelia:
> Realizing that one issue with this design is that we are setting the same vertex buffer over and over for each phase item. Wondering if we should either:

Actually, this appears to be a no-op, so we're fine! https://github.com/bevyengine/bevy/blob/main/crates/bevy_render/src/render_phase/draw_state.rs#L191-L202

bevy_nannou_render/src/pipeline.rs
```diff
@@ -74,14 +62,19 @@ impl NannouPipeline {
...
         &self,
         color_format: wgpu::TextureFormat,
-        depth_format: wgpu::TextureFormat,
         sample_count: u32,
         blend_state: wgpu::BlendState,
         topology: wgpu::PrimitiveTopology,
     ) -> RenderPipelineDescriptor {
+        let view_key = MeshPipelineKey::from_msaa_samples(sample_count)
+            | MeshPipelineKey::from_hdr(color_format == ViewTarget::TEXTURE_FORMAT_HDR)
+            | MeshPipelineKey::from_primitive_topology(topology);
```
mitchmindtree, line 71: Oooo fancy :star_struck:

result
```diff
@@ -0,0 +1 @@
+/nix/store/l6qwc0cii9g6hzrqcb79wr61jg32nhx6-nannou-0.19.0
```
mitchmindtree, line 1: Looks like `result` snuck in here :sweat_smile:

mitchmindtree (approved):
@tychedelia all I can say is - this is looking awesome! Seems like someone is becoming a bevy pro :dove: :nerd_face: 

> Please bikeshed names

Honestly I think I don't mind it too much, as it kinda makes sense when reading those commands in context :thinking:

> In bevy_nannou_render/src/lib.rs, notice how much code was able to be removed from update_draw_mesh (previously prepare_view_mesh.

Huge :rocket: In general, it's so cool to see the draw logic broken up into bevy systems.

> In particular, it's still worth investigating whether we can make more use of Bevy's mesh infrastructure, rather than re-inventing the wheel here

So curious about this - would be amazing if we could leverage bevy's mesh abstractions too. Have you had a chance to play with bevy's mesh stuff yet / get an intuition for how feasible this might be?

tychedelia:
Going to merge this and begin working on the follow up outlined here: https://github.com/nannou-org/nannou/issues/966

PR #974, opened 2024-07-31
Add new render function for raw wgpu.

tychedelia:
Adds a new render function type to the app builder that allows access to raw `wgpu` resources and restores the `Frame` concept for this function.

```rust
fn main() {
    nannou::app(model).render(render).run();
}

fn render(app: &RenderApp, model: &Model, frame: Frame) {
    // do something with `wgpu`
}
```

We originally had to remove `Frame` as it no longer made sense in Bevy's pipelined render model. In Bevy, `wgpu` resources are held in the render world, which means that we couldn't provide access to them in our default `update`, `view`, etc., functions. Additionally, because we now render into Bevy's `Mesh` component, there was no need to access these lower level `wgpu` concepts in `view`.

However, some users may still want to manually wire everything up themselves, and for this, we now provide `render`.

One major change here is the addition of a `RenderApp` type that mirrors `App`. Not all methods from `App` have been migrated here and due to the complexities of accessing the same data in the render world, we'll want to check to see what makes sense.

One caveat that I'm still a little nervous about is that Bevy provides a lot of utilities for stuff like creating bind groups, etc., that our users could really benefit from. Adding this `render` fn is a good bridge to minimize breaking changes for our existing users, but part of me still feels like new users could benefit from the reduced boilerplate offered by Bevy. More to consider later.

Notable changes:
- To use `render` your model must have a `M: Clone` bound in order for it to be cloned into the render world. For the most part, this means putting `Arc` around any `wgpu` resources.
- Because we are using a Bevy render node to place our `wgpu` operations in the context of Bevy's broader pipeline, users who wish to use msaa need to target the resolve attachment rather than the intermediate texture view.

How to review this:

Mostly you should look at the changes in `nannou/src` that show how we are wiring this into Bevy. The other changes are mostly just adding back the `nannou_wgpu` crate and minor changes to the examples to get them to compile. Those examples can be manually diffed against `master` in a local checkout to see what changed.

image


tychedelia:
@MacTuitui this probably is too chaotic to review, but you may be interested in this one as it should drastically ease your migration onto the bevy version if you choose to do so. Let me know if you have any questions or concerns.

PR #989, opened 2025-02-08
fix: Get nannou package derivation and devShell building on macOS

mitchmindtree:
The SDK packages and libiconv have since been included in the nix stdenv for darwin builds. Removing them gets the nannou package building again on macOS 🍎 

cc @supiket

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -16,7 +16,12 @@
 }:
 rustPlatform.buildRustPackage rec {
   pname = "nannou";
-  src = ./.;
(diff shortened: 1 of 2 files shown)


supiket:
Confirming that the following steps led to a successful build (thanks @mitchmindtree!):
> `nix shell nixpkgs#lld nixpkgs#darwin.libiconv`
> `RUSTFLAGS="--cfg=web_sys_unstable_apis -C linker=wasm-ld" cargo build -p nannou --target wasm32-unknown-unknown`

🤯

PR #991, opened 2025-02-11
deps: propagate bevy feature 'webgl2'

supiket:
This is seemingly required when building for `wasm32-unknown-unknown`


supiket:
The necessary bevy feature seems to be wgpu, which will end up in `bevy-refactor` branch soon

Issue #1003, opened 2025-03-04
Complete `draw.text()` implementation


mitchmindtree:
For now, the `draw::render::RenderPrimitive` implementation for `Text` is still `todo!()`.

PR #1007, opened 2025-03-04
fix: Address `bevy_nannou_draw` warnings

mitchmindtree:
I thought I'd have a go at cleaning up some of the warnings first :nerd_face: Starting one crate at a time.

cc @tychedelia do you happen to know off hand if we need the `SRes<RenderAssets<GpuShaderStorageBuffer>>` field in `RenderCommand::Param` for `DrawMeshInstanced` anymore? Just double checking I'm not removing something important :sweat_smile: 

I've opened an issue to track the `todo!` work for text here at https://github.com/nannou-org/nannou/issues/1003.

Also I've just removed the `texture_handle` from `Path` for now - I figure we can re-add if we remember what it was for (it might have just been a hack to keep the texture from dropping in a previous version or something?).


tychedelia:
> cc @tychedelia do you happen to know off hand if we need the SRes<RenderAssets<GpuShaderStorageBuffer>> field in RenderCommand::Param for DrawMeshInstanced anymore? Just double checking I'm not removing something important 😅

I think this was in early attempts to provide a way for the user to override the vertex buffer for an instanced draw call to implement vertex pulling. Thanks so much for cleaning this up!

PR #1016, opened 2025-03-06
fix: change `bevy_nannou_derive` to only expect the presence of `bevy_nannou`, not the `nannou` crate

### LINKED ISSUE #999: `bevy_nannou_derive` ideally usable without `nannou` (only `bevy_nannou`)
As discussed, this is probs an easy fix!

mitchmindtree:
Closes #999


mitchmindtree:
Ahh I just realised this means that `bevy_nannou` needs to be in the user's `Cargo.toml`, and that if you only have `nannou` it's not enough :thinking: I think this option is better of the two though, as `bevy_nannou` is the most upstream which minimizes the required deps to pull in.

PR #1020, opened 2025-03-16
fix: Lazily initialize Mesh in draw commands for webgl2 compat

### LINKED ISSUE #998: Allow targeting WebGL using `webgl2` bevy feature
Related to #997, would be nice to have the option for targeting web with greater compatibility.

mitchmindtree:
WebGL appears to have a requirement that initialised buffers must have a non-zero length (or something along these lines). This changes our draw logic to lazily initialise the mesh to work around this requirement.

Closes #998.

cc @tychedelia does this solution look OK?

### DIFF
--- a/bevy_nannou_draw/src/render.rs
+++ b/bevy_nannou_draw/src/render.rs
@@ -584,7 +584,7 @@ fn update_draw_mesh(
         let mut stroke_tessellator = StrokeTessellator::new();
 
         let mut last_shader_model = None;
-        let mut mesh = meshes.add(Mesh::init());
+        let mut current_mesh = None;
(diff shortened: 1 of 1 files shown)


tychedelia (approved):
Yes! This totally makes sense and is definitely the correct implementation with respect to this. The "mono mesh" is potentially a weird optimization but seemed to work pretty well. If we wanted in the future, it might make sense to optimize further by retaining vertex buffers, but in general these kinds of optimizations are really hard to do with the imperative drawing model as the order and number of draw commands can be totally arbitrary frame to frame. Tysm for tracking this down!

PR #1032, opened 2025-11-09
Fix macOS build error by updating hotglsl to use naga

ghaworth:
## Description
Fixes build failures on macOS caused by the outdated `glsl-to-spirv` dependency.

## Problem
The current version of `hotglsl` (commit 1c1a303) uses `glsl-to-spirv` v0.1.7, which fails to compile on modern macOS systems with the error:
```
fatal error: 'unordered_map' file not found
```

This prevents nannou from building on Apple Silicon Macs and other modern macOS systems.

## Solution
- Updated `hotglsl` dependency to latest commit (534b228) which uses `naga` instead of `glsl-to-spirv`
- Updated `nannou_isf/src/pipeline.rs` to use `ShaderStage::Fragment` instead of the deprecated `ShaderType::Fragment` to match hotglsl's new API

## Testing
- Tested on macOS with Apple Silicon (M-series chip)
- Successfully builds with `cargo build --release`
- Successfully runs `simple_window` example

## Related
This aligns with the upstream hotglsl migration from glsl-to-spirv to naga (hotglsl commit 005aa24: "Rm old glsl-to-spirv crate in favour of naga").

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -409,16 +409,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0d8c1fef690941d3e7788d328517591fecc684c084084702d6ff1641e993699a"
 
 [[package]]
-name = "block-buffer"
-version = "0.3.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a076c298b9ecdb530ed9d967e74a6027d6a7478924520acddcddc24c1c8ab3ab"
-dependencies = [
- "arrayref",
- "byte-tools",
-]
-
-[[package]]
 name = "block-sys"
 version = "0.1.0-beta.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


JoshuaBatty:
Thanks!

ghaworth:
Hi Joshua,

I noticed the CI tests are failing, but the error appears to be unrelated to my changes. The failure is in `cargo` v0.42.0 which has compatibility issues with newer Rust versions:
```
error[E0308]: mismatched types
  --> cargo-0.42.0/src/cargo/util/flock.rs:241:49
```

This seems to be a pre-existing issue in the test infrastructure (possibly in `scripts/run_all_examples/Cargo.toml` which shows a warning about defaulting to 2015 edition).

My changes only touch:
- `Cargo.lock` (updating hotglsl)
- `nannou_isf/src/pipeline.rs` (ShaderType → ShaderStage)

I've tested the build successfully on macOS. Please let me know if you'd like me to investigate the CI issue or if this can be merged as-is since the failure is unrelated to the fix.

Thanks,
George

PR #1035, opened 2026-02-23
Depend on cpal 0.14 for more robust ALSA support

sooda:
See https://github.com/RustAudio/cpal/issues/606. Fixes at least examples simple_audio and simple_audio_file that might fail like so:

    thread 'main' (26045) panicked at examples/audio/simple_audio_file.rs:35:10:
    called `Result::unwrap()` on an `Err` value: BuildStream { err: BackendSpecific { err: BackendSpecificError { description: "ALSA function 'snd_pcm_sw_params' failed with error 'EINVAL: Invalid argument'" } } }

Nothing else tested comprehensively but at least cargo build on the examples is happy with the move from cpal 0.13.

Also consider a patch version update to have the fix available.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2974,7 +2974,7 @@ dependencies = [
 
 [[package]]
 name = "nannou_audio"
-version = "0.19.0"
+version = "0.19.1"
 dependencies = [
  "cpal",
  "dasp_sample",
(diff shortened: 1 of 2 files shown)


mitchmindtree:
Thanks for this, and sorry for the delay! I've rebased your commits and extended with a more comprehensive update in #1069. Edit: I'll aim to publish it along with the rest of nannou under 0.20 when the new bevy release drops.

PR #1060, opened 2026-06-09
Update noise dependency to 0.9

### LINKED ISSUE #1050: Update or replace the `noise` dependency
`noise` `0.7.0` defines `Perlin` in both `perlin` and `perlin_surflet`, so `noise::Perlin` is ambiguous - now a hard error on recent rustc, which we worked around with `[lints] ambiguous_glob_imports = "allow"` on the examples/generative_design packages.

Bump `noise` (`0.9` has API churn) or migrate off it then drop the lint allows.

ss123she:
Closes #1050

Updates noise dependency to 0.9 and drops the now redundant `ambiguous_glob_imports = "allow"` lints.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -79,12 +79,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "adler"
-version = "1.0.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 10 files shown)


mitchmindtree:
Thanks !

PR #1075, opened 2026-06-21
fix(nannou): compile pipelines synchronously so the persistent canvas works from frame `0`

mitchmindtree:
nannou cameras use 4x MSAA by default, and Bevy's no-clear persistence with MSAA relies on the `msaa_writeback` blit pass to carry the previous frame forward into the multisampled target each frame.

That blit pipeline compiles asynchronously over ~40 frames on a cold cache - until it is ready nothing persists, so anything drawn during that window - e.g. a sketch that composes its whole image once on frame 0 - was silently dropped. Several Generative Gestaltung examples (`p_1_2_3_02`, `p_1_2_3_04`, ...) rendered black as a result.

Default Bevy's `RenderPlugin::synchronous_pipeline_compilation` to `true` so the writeback pipeline is ready before the first frame. Add `Builder::synchronous_pipeline_compilation` to opt back into asynchronous compilation; plugin setup is deferred out of `Builder::new` into a lazy `ensure_initialized` guard so the flag can be configured beforehand.

@tychedelia does this look like an OK solution? I'm wondering if there's a way to keep async pipeline compilation by default but gate nannou's first frame until the pipeline is ready or something along these lines :thinking:

### DIFF
--- a/nannou/src/app.rs
+++ b/nannou/src/app.rs
@@ -107,6 +107,13 @@ pub struct Builder<M = (), E = WindowEvent> {
     render: Option<RenderFn<M>>,
     default_view: Option<View<M>>,
     exit: Option<ExitFn<M>>,
+    /// Whether render pipelines compile synchronously (the default). See
+    /// [`Builder::synchronous_pipeline_compilation`].
+    synchronous_pipeline_compilation: bool,
+    /// Whether the default Bevy plugins have been added to `app` yet. Plugin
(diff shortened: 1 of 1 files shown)


tychedelia (approved):
This is basically what we're doing for processing. Unfortunately there's not a great way right now in Bevy to check if pipelines are ready. If you search the issue tracker you can probably find some references to this.

PR #1096, opened 2026-08-30
fix(draw): close the ellipse outline path so strokes don't leave a gap

### LINKED ISSUE #1095: Ellipses aren't closed
It seems that the `.ellipse` method doesn't close the arc that's used in the non-resolution branch. That is, this results in a closed (approximated) circle:
```rust
    draw.ellipse()
        .stroke_weight(100.0)
        .stroke(BLACK)
        .w_h(200.0, 200.0)
        .resolution(128.0);
```
While this produces an unclosed arc:
```rust
    draw.ellipse()
        .stroke_weight(100.0)
        .stroke(BLACK)
        .w_h(200.0, 200.0);
```

I believe the fix is simply to add `builder.close()` before `builder.build()` in [`ellipse.rs`](https://github.com/nannou-org/nannou/blob/bffa9887b1858c5f84fdde9560f00412b5013489/nannou_draw/src/draw/primitive/ellipse.rs#L84). (Note that the `match` branch *with* a resolution does pass `true` to the `close` parameter of `polygon::render_points_themed`.)

See the following figure for an illustration of the problem:

Image

(A secondary point is that when one *does* supply a resolution, the corresponding match arm adds a point to the path for each element of the `Ellipse::circumference` iterator. This produces a `Circumference` with the given resolution, which again calls `Circumference::new_inner` with `num_points` set to `resolution + 1` and a `rad_step` of `TAU / resolution`. Unless I'm mistaken, this means that the first and last points (approximately) coincide, meaning that the closing of the path yields a (close-to-)zero-length segment. Might not be worth addressing, since using `num_points = resolution` dosn't seem to be right for *all* consumers, here, and since Lyon will presumably filter out the degeneracy?)

Jorge-Polanco-Roque:
**Problem:** `draw.ellipse()` without `.resolution(...)` builds an open arc, so a stroked ellipse leaves a gap where the path's start and end caps meet instead of joining cleanly (#1095). The `.resolution(...)` path already closes.

**Fix:** close the outline path before building it. The path construction is pulled into a small `ellipse_outline_path` helper so it can be unit-tested; the resolution branch is unchanged (its near-zero-length closing segment is filtered by lyon, as the reporter noted).

**Tests:** added `ellipse_outline_path_is_closed` (asserts the path emits `PathEvent::End { close: true }`) and `degenerate_ellipse_has_no_outline`; verified the first fails without the fix. `cargo fmt`, `clippy`, and the `nannou_draw` tests are green.

Closes #1095


Jorge-Polanco-Roque:
Gentle ping — this has been open about nine days. It closes the ellipse outline path so strokes don't leave a visible gap. Happy to rebase or adjust anything whenever a maintainer can take a look. Thanks for nannou!

Jorge-Polanco-Roque:
Going to close this for now to keep my open PRs tidy — no worries on your end, the repo's clearly had other priorities lately. The fix still stands (closing the ellipse outline path so strokes don't leave a gap at the seam); happy to reopen or resubmit whenever it's useful. Thanks for nannou!

ffunatsu:
@Jorge-Polanco-Roque 
I'm not a collaborator, but I personally checked this PR and the Issue, just in my environment, it reproduced on the current latest nannou and fixed by this PR.

This PR code itself seems nice, so I think this PR should be reopened.

